PR #1, opened 2020-05-15
Add import statements to preview and improve icon-details navigation

peterhass:
https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-29157

### DIFF
--- a/packages/icons/preview/import_statements.jsx
+++ b/packages/icons/preview/import_statements.jsx
@@ -13,11 +13,11 @@ const getAlias = (iconName, suffix, startUpper) => {
 export const importStatementForReactComponent = (size, fileName) => {
   const name = getIconName(fileName);
   const alias = getAlias(name, 'Icon', true);
-  return `import { ReactComponent as ${alias} } from "@goodhood/icons/svg/${size}/${fileName}"`;
+  return `import { ReactComponent as ${alias} } from '@goodhood/icons/svg/${size}/${fileName}';`;
 };
 
 export const importStatementForImg = (size, fileName) => {
   const name = getIconName(fileName);
   const alias = getAlias(name, 'Icon', false);
-  return `import ${alias} from "@goodhood/icons/svg/${size}/${name}"`;
+  return `import ${alias} from '@goodhood/icons/svg/${size}/${name}';`;
 };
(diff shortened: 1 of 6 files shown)


packages/icons/preview/import_statements.jsx
```diff
@@ -0,0 +1,16 @@
+import upperFirst from 'lodash/upperFirst';
+import camelCase from 'lodash/camelCase';
+
+const getIconName = (fileName) => fileName.substr(0, fileName.lastIndexOf('.'));
```
suprMax, line 4: `fileName.replace(/\.svg$/, '')` bit simpler?

packages/icons/preview/import_statements.jsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const importStatementForReactComponent = (size, fileName) => {
+  const name = getIconName(fileName);
+  const alias = upperFirst(camelCase(`${name}_icon`));
```
suprMax, line 8: I think the name is wrong, needs to be `WhateverIcon` not `Whatever_icon`.

packages/icons/preview/import_statements.jsx
```diff
@@ -0,0 +1,16 @@
...
+  return `import { ReactComponent as ${alias} } from "@goodhood/icons/svg/${size}/${fileName}"`;
+}
+
+export const importStatementForImg = (size, fileName) => {
+  const name = getIconName(fileName);
+  const alias = camelCase(`${name}_icon`);
```
suprMax, line 14: i think in this case would make sense `whateverUrl` not `whatever_icon`

packages/icons/preview/import_statements.jsx
peterhass: result is the same: 
'whatever_icon` -> 'whateverIcon' -> 'WhateverIcon'

suprMax: sorry my bad, didn't notice you pass that in

peterhass: Yes but it's a good sign that it's somehow confusing. Will add the 'Icon' part after upperFirst and camelCase

packages/icons/preview/import_statements.jsx
peterhass: Imo it would improve readability if the import alias has the word 'icon' in it.

WhateverIcon - React Component rendering the Icon
whateverIcon - URL to Icon

imo the information that it's a icon is more important than the data behind the alias. As soon as I see a icon with lowercase starting letter, I know it's a url.

In nebenan-frontend we already imported some svgs and we **never** applied the 'Url' suffix.

peterhass: Screenshot 2020-05-15 at 18 29 19

IconUrl suffix is weird too

PR #2, opened 2020-05-18
Fix icons

peterhass:


peterhass:
Screenshot 2020-05-18 at 12 02 50
svgr-loader converts those rects into `path` elements. I need to know which element to apply fill-color to - therefore it would be best to not have those empty rects

peterhass:
Found a css side solution. Let's discuss in nebenan-frontend https://github.com/good-hood-gmbh/nebenan-frontend/pull/1027 PR

PR #4, opened 2020-06-15
Components

peterhass:
https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-29566

- Used in nebenan-static-errors https://github.com/good-hood-gmbh/nebenan-static-errors/pull/2


package.json
```diff
@@ -17,6 +17,7 @@
   },
   "scripts": {
     "lint": "lerna run lint",
-    "postinstall": "lerna bootstrap"
+    "postinstall": "lerna bootstrap",
+    "build": "rollup --config"
```
suprMax, line 21: does root repo has/need a build?

packages/components/.babelrc
```diff
@@ -0,0 +1,23 @@
+{
+  "presets": [
+    ["@babel/preset-env", {"modules": false}],
```
suprMax, line 3: modules should be probably 'auto' no?

packages/components/.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
+lib/
```
suprMax, line 2: EOF line

packages/components/.npmignore
```diff
@@ -0,0 +1,7 @@
+.storybook/
+src/
+.babelrc
+.eslintrc
```
suprMax, line 4: mocha config can be ignored too. everything that we don't need to import this package can/should be ignored

packages/components/.storybook/main.js
```diff
@@ -0,0 +1,52 @@
+const sassFunctions = require('sass-functions');
+const sass = require('sass');
+const sassGlobImporter = require('node-sass-glob-importer');
```
suprMax, line 3: is this needed?

packages/components/README.md
```diff
@@ -0,0 +1,5 @@
+# `@goodhood/components`
+
+> A collection of React Nebenan UI Components
+
+## Usage
```
suprMax, line 5: could write a bit here

packages/components/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "date-fns": "^2.14.0",
+    "eslint": "^6.6.0",
+    "eslint-config-nebenan": "^1.3.2",
+    "lodash": "^4.17.15",
+    "mocha": "^8.0.1",
+    "nebenan-helpers": "^4.2.0",
+    "nebenan-ui-kit": "^4.12.0",
+    "node-sass-glob-importer": "^5.3.2",
+    "prop-types": "^15.7.2",
+    "qs": "^6.9.4",
```
suprMax, line 59: drop unused deps

packages/components/rollup.config.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    {
+      file: pkg.main,
+      format: 'cjs',
+      sourcemap: true,
+    },
+  ],
+  plugins: [
+    del({
+      targets: path.join(pkg.module, '../*'),
```
suprMax, line 30: what is this deleting exactly?

packages/components/rollup.config.js
```diff
@@ -0,0 +1,53 @@
...
+    }),
+    peerDepsExternal(),
+    postcss({
+      extract: true,
+      modules: true,
```
suprMax, line 35: don't we need the classnames magic here same as in the core repo?

packages/components/rollup.config.js
```diff
@@ -0,0 +1,53 @@
...
+      use: [['sass', {
+        includePaths: [
+          `${__dirname}/node_modules`,
+        ],
+        functions: sassFunctions({ sass }),
+        importer: sassGlobImporter(),
```
suprMax, line 41: is this needed?

packages/components/rollup.config.js
```diff
@@ -0,0 +1,53 @@
...
+        outputStyle: 'expanded',
+        sourceComments: true,
+      }]],
+    }),
+    babel({
```
suprMax, line 46: does this use .babelrc?

packages/components/src/index.jsx
```diff
@@ -0,0 +1,3 @@
+export { default as Dots } from './dots';
```
suprMax, line 1: would be better to autogenerate this file. ideas?

packages/components/.storybook/main.js
```diff
@@ -0,0 +1,52 @@
...
+        modules: modules,
+      }
+    },
+    {
+      loader: 'sass-loader',
+      options: {
+        implementation: sass,
+
+        sassOptions: {
+          includePaths: [
```
peterhass, line 20: same options as in rollup.config.js but not sure the effort to refactor (cfs <-> esm)

package.json
peterhass: No, shouldn't be there. Will remove that line

packages/components/rollup.config.js
peterhass: this is cleaning up the lib directory, better use `__dirname` and `'lib/*'` ?

packages/components/rollup.config.js
peterhass: yes

packages/components/rollup.config.js
peterhass: No I wouldn't use the long fancy class names for the finished components build. They can't be overwritten anyway. Would rather spend time to keep component names in devtools

Preview is something different - there it could be useful.

packages/components/src/index.jsx
peterhass: I used a plugin to autogenerate those imports. As long as every folder exports one default-export - it's easy but Components like MicroHelmet export default and Provider

Therefore I switched over to the manual approach for now.  Any idea on how to handle these cases?

packages/components/.babelrc
peterhass: https://babeljs.io/docs/en/babel-preset-env#modules

False means: Keep ESM module syntax

Otherwise babel would transform all `import blubb from 'blabb'` to `requre(...)` -> no treeshaking

packages/components/rollup.config.js
peterhass:
```
      targets: path.join(__dirname, 'lib/*'),
```

packages/components/.babelrc
suprMax: i thought it adjusts for rollup automagically

packages/components/rollup.config.js
suprMax: clearer yes

packages/components/src/index.jsx
suprMax: `export * as camelCase, { default as SnakeCase } from 'whatever';`

would something like this work? (play around with syntax, this is probably incorrect)

packages/components/.mocharc.js
```diff
@@ -0,0 +1 @@
+module.exports = require("../../presets/react/mocha.config.js");
```
peterhass, line 1: `extends: ` didn't work for me. It applied things like `require` but not `reporter` or `ui`

packages/components/.storybook/manager.js
```diff
@@ -0,0 +1,5 @@
+import { addons } from '@storybook/addons';
+import setup from '../../../presets/react/storybook/manager';
+import pkg from "../package.json";
+
+setup({ pkg, addons });
```
peterhass, line 5: addons holds some kind of store, if I import it in presets/react/storybook/manager, the store will be different to the one used in this package

packages/components/.storybook/preview.js
```diff
@@ -0,0 +1,4 @@
+import * as clientApi from "@storybook/client-api";
+import setup from '../../../presets/react/storybook/preview';
+
+setup({ clientApi });
```
peterhass, line 4: same as here https://github.com/goodhood-eu/goodhood/pull/4/files#r441476620

.browserslistrc
```diff
@@ -0,0 +1,7 @@
+# Browsers that we support. To see current list: `npx browserslist`
+
+last 2 major versions
+> 1%
```
suprMax, line 4: copy current from core, there is also ie11 drop

packages/components/.mocharc.js
suprMax: probably same for every package, not react specific

packages/components/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "test": "NODE_ENV=test TZ='GMT' mocha \"./{,!(node_modules)/**}/*.test.{js,jsx}\"",
+    "test:tofile": "NODE_ENV=test TZ='GMT' mocha --no-colors -R spec \"./{,!(node_modules)/**}/*.test.{js,jsx}\" 2>&1 | tee test-report.log"
+  },
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "clsx": "^1.1.1",
+    "date-fns": "^2.14.0",
+    "nebenan-helpers": "^4.2.0",
+    "prop-types": "^15.7.2",
+    "react": "^16.13.1",
```
suprMax, line 34: move to upper devdeps?

packages/components/src/index.jsx
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/no-unresolved
+export * from './*/index.jsx';
```
suprMax, line 2: is this magic expanded by rollup? leave a comment please

PR #5, opened 2020-06-19
SSR

peterhass:
Issues with this implementation:
- Need to use storybooks internal webpack config and patch it (may change in future versions)

Known issues:
- Open story, edit source to throw an error, error message is shown, revert source changes, updated component is shown but error message is still there (not sure if it's worth the effort to fix it, would need a error boundary)


packages/components/package.json
```diff
@@ -22,6 +22,7 @@
     "lint": "eslint -f codeframe \"./**/*.{js,jsx}\"",
     "build": "rollup --config rollup.config.js",
     "start": "start-storybook -p 3000 --ci",
+    "start:ssr-preview": "SSR=true start-storybook -p 3000 --ci --preview-url=/ssr-iframe",
```
peterhass, line 25: suggestion on how to extract this part? 

maybe `node ../../presets/react/scripts/start-storybook-ssr.js` ?

packages/components/.storybook/middleware.js
```diff
@@ -0,0 +1,8 @@
+const path = require('path');
+
+const pkgPath = path.join(__dirname, '../');
+const getMiddleware = require('../../../presets/react/storybook/middleware');
+
+module.exports = (router) => {
```
suprMax, line 6: i'm not familiar with .storybook enough but couldn't we point local storybook server to a configuration folder a few levels above? this dancing around with importing/extending configs would be avoided then?

packages/components/src/micro_helmet/index.stories.jsx
```diff
@@ -3,7 +3,7 @@ import MicroHelmet from './index';
 import MicroHelmetProvider from './provider';
 
 const usePageTitle = () => {
-  const [title, setTitle] = useState(document.title);
+  const [title, setTitle] = useState(typeof document !== 'undefined' && document.title);
```
suprMax, line 6: default title value is `false`, is that what you wanted?

presets/react/storybook/middleware.js
```diff
@@ -0,0 +1,11 @@
+const { Router } = require('express');
+
+module.exports = ({ pkgPath }) => {
+  const router = Router();
+
+  if (process.env.SSR === 'true') {
```
suprMax, line 6: since the whole point of this middleware is to provide SSR the check seems unnecessary

presets/react/storybook/preview-ssr.js
```diff
@@ -0,0 +1,3 @@
+import setup from './preview';
```
suprMax, line 1: what is the purpose of this file?

presets/react/storybook/ssr/src/utils.jsx
```diff
@@ -0,0 +1,17 @@
+import qs from 'qs';
+
+export const getStories = () => {
+  const stories = require.context(`${PKG_PATH}/src`, true, /\.stories\.jsx$/);
+  return stories.keys().map(stories);
+};
+
+export const getUrlForStory = ({ storyId, viewMode }) => {
+  const { path, selectedKind, selectedStory, ...rest } = qs.parse(document.location.search, {
+    ignoreQueryPrefix: true,
```
suprMax, line 10: slicing would be shorter code

presets/react/storybook/ssr/webpack.config.js
```diff
@@ -0,0 +1,162 @@
...
+const { DefinePlugin, HotModuleReplacementPlugin } = require('webpack');
+const babelConfig = require('../../babel.config');
+
+const CONFIG_NAME_CLIENT = 'client';
+const CONFIG_NAME_SERVER = 'server';
+const COMPILE_INDEX_SERVER = 0;
+const COMPILE_INDEX_CLIENT = 1;
+const ROOT_PKG_PATH = path.resolve(path.join(__dirname, '../../../../'));
+
+const fileLoaderRules = [{
```
suprMax, line 13: rewriting this feels super iffy, wouldn't it be possible to import webpack config from storybook and mutate the parts that we need?

presets/react/storybook/middleware.js
peterhass: `.storybook/middleware.js` is a configuration file from storybook.

- storybook loads packages/components/.storybook/middleware.js 
- always loads `presets/react/storybook/middleware.js`
- then decides if it should use SSR middleware or not

Wouldn't want to move the `env.SSR` check into package's `middleware.js`

presets/react/storybook/preview-ssr.js
peterhass: storybook loads preview.js when running the preview in the browser. it is used for preview configuration and initialisation (you could initialize stories here too). I splitted preview.js from preview-ssr.js because there are some limitations in ssr preview. 
For example: You cannot load additional stories. We would need them on server side too.

That's why I want subpackages to have specific configs for preview and ssr preview. Would lead to errors that are hard to debug if you don't know all internals of storybook

presets/react/storybook/middleware.js
suprMax: we always need it though? when is the case we don't want SSR

peterhass: Wouldn't want SSR during normal development. It is way way slower, it has to do full refreshs all the time and the error handling is way nicer on the client. Would rather develop without SSR and check afterwards if it works with SSR enabled

packages/components/src/micro_helmet/index.stories.jsx
peterhass: Yes

presets/react/storybook/ssr/src/utils.jsx
peterhass: used nebenan-helpers instead, should be better now

presets/react/storybook/ssr/src/utils.jsx
```diff
@@ -0,0 +1,15 @@
+import { getSearch, getQuery } from 'nebenan-helpers/lib/routes/';
+
+export const getStories = () => {
+  const stories = require.context(`${PKG_PATH}/src`, true, /\.stories\.jsx$/);
```
suprMax, line 4: i don't see where PKG_PATH is defined

peterhass: `DefinePlugin` in webpack config

PR #8, opened 2020-06-25
Pre-built icons

peterhass:
https://jira.nebenan.de/browse/CORE-5441


packages/icons/preview/icons.jsx
```diff
@@ -1,8 +1,15 @@
 const tree = require('./icon_list');
 
+const getAssets = (size, name) => {
+  const { ReactComponent } = require(`../src/${size}/${name}?react`);
+  const { default: url } = require(`../src/${size}/${name}`);
```
peterhass, line 5: loads raw svg through svgo-loader and the react component through @svgr/webpack

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,93 @@
...
+      path.basename(svgPath).replace(/\.svg$/, ''),
+    ),
+  )
+);
+
+const getFileName = (svgPath) => (
+  snakeCase(
+    path.basename(svgPath).replace(/\.svg$/, ''),
+  )
+);
```
peterhass, line 57: almost the same as in `import_statements.jsx` - extract somehow?

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,93 @@
...
+};
+
+const getFiles = (tree) => (
+  Object.keys(tree)
+    .reduce((acc, size) => (
+      acc.concat(tree[size].map((name) => (
+        `${size}/${name}`
+      )))
+    ), [])
+);
```
peterhass, line 43: copied from `import_statements.jsx` - suggestion where to extract this stuff to?

packages/icons/svgr.config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  svgo: true,
+  plugins: ['@svgr/plugin-svgo', '@svgr/plugin-jsx'],
```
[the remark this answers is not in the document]
peterhass: I just tried out combining svgr/webpack and svgo-loader loaders and found a big issue:

@svgr/webpack doesn't use previous results in the loader chain. It always uses the raw resource file. Therefore we cannot pipe svgo-loader results into it. @svgr/webpack has to call svgo on it's own.

In short: We have to keep this line as it is.

packages/icons/scripts/build.js
```diff
@@ -0,0 +1,46 @@
...
+
+  console.log(`${svgPath}\n--> ${chalk.magenta(libSvgPath)}\n--> ${chalk.red(libReactPath)}\n`);
+
+  const data = fs.readFileSync(svgPath, 'utf-8');
+
+  const [
+    { data: optimizedData },
+    reactComponentCode,
+  ] = await Promise.all([
+    svgo.optimize(data, { path: svgPath }),
```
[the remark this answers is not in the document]
peterhass: https://github.com/goodhood-eu/goodhood/pull/8#discussion_r447493388

> Wasn’t able to use svgo-loader in combination with @svgr/webpack - because of that reason, I think we should not change the build script to pass the optimized svg to svgr because it would complicate svgr config (we would need two different configs - one for webpack, one for build)

PR #9, opened 2020-06-26
Map package

atomiomi:


packages/map/README.md
```diff
@@ -0,0 +1,4 @@
+nebenan-map
```
peterhass, line 1: `@goodhood/maps`

packages/map/README.md
```diff
@@ -0,0 +1,4 @@
...
+============
+
+React map components
```
peterhass, line 4: what about a small 'usage' section?

packages/map/src/index.jsx
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/no-unresolved */
+
+// glob gets magically expanded by rollup
+export * from './*/index.jsx';
```
peterhass, line 4: Do you really want this? This line exports stuff like utils too

presets/react/storybook/preset.js
```diff
@@ -42,6 +42,8 @@ module.exports = {
     '@storybook/addon-viewport/register',
     '@storybook/addon-storysource',
     '@storybook/addon-docs',
+    '@storybook/addon-knobs/register',
```
peterhass, line 45: do they work with SSR enabled?

packages/map/.npmignore
```diff
@@ -0,0 +1,11 @@
+.storybook/
```
peterhass, line 1: https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d

using package.json `files` may be simpler?

packages/map/.gitignore
```diff
@@ -0,0 +1,4 @@
+styles.css
+lib
+test-report.log
+sample_data/local.jsx
```
peterhass, line 4: can we get the app to run without this file too? (so that it doesn't fail on first `npm start`)

packages/map/README.md
```diff
@@ -0,0 +1,4 @@
+nebenan-map
+============
+
```
peterhass, line 3: how do I get the preview to run?

packages/map/src/map/utils.jsx
```diff
@@ -0,0 +1,23 @@
...
+  if (!isFilledArray(bounds)) return undefined;
+
+  // Reference: https://docs.mapbox.com/mapbox-gl-js/example/zoomto-linestring/
+  const initial = new LngLatBounds(bounds[0], bounds[0]);
+  const lngLat = bounds.reduce((acc, item) => acc.extend(item), initial);
+
+  return lngLat.toArray();
+};
+
+export const mergeChildrenBounds = (bounds) => {
```
suprMax, line 20: a couple tests for utils?

packages/map/src/marker/utils.jsx
```diff
@@ -0,0 +1,18 @@
+import { render, unmountComponentAtNode } from 'react-dom';
+import { Popup } from 'mapbox-gl';
+
+
+export const createPopup = (win, content, options) => {
+  const element = win.document.createElement('div');
```
suprMax, line 6: any reason not to render popup content as marker child node and also pass ref?

packages/map/src/pin_marker/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+  PIN_MARKER_BASE,
+  PIN_MARKER_ORANGE,
+  PIN_MARKER_YELLOW,
+  PIN_MARKER_RED,
+  PIN_MARKER_GRAY,
+  PIN_MARKER_BLUE,
+];
+
+const images = types.reduce((acc, type) => {
+  acc[type] = require(`./images/pin-${type}.svg`);
```
suprMax, line 28: use glob import instead?

packages/map/src/pin_marker/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+  return acc;
+}, {});
+
+const PinMarker = ({
+  children,
+  type,
+  ...rest
+}) => (
+  <Marker {...rest}>
+    <img src={images[type]} alt="" />
```
suprMax, line 38: set alt to `type`?

packages/map/src/polygon/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+  const {
+    type,
+    area,
+    onClick,
+  } = props;
+
+  useChildrenBounds(area);
+
+  const sourceId = useRef(getID()).current;
+  const layerId = useRef(getID()).current;
```
suprMax, line 19: tbh still looks like a package level hook would make sense? instead of import `getID` could import `useUniqueId()`

presets/react/storybook/preset.js
atomiomi: works

packages/map/sample_data/index.jsx
```diff
@@ -0,0 +1,49 @@
+import local from './local';
```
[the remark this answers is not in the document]
atomiomi: maybe in another PR. Will need help from Peter

packages/map/src/image_marker/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+
+const ImageMarker = ({
+  children,
+  image,
+  caption,
+  ...rest
+}) => (
+  <Marker {...rest}>
+    <span className={styles.container} style={{ backgroundImage: `url(${image})` }} />
```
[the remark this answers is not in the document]
atomiomi: not all passed images will be with 1:1 ratio.

packages/map/src/map/hooks.jsx
```diff
@@ -0,0 +1,64 @@
...
+      doubleClickZoom: false,
+      scrollZoom: false,
+      dragRotate: false,
+      pitchWithRotate: false,
+      apiUrl: null,
+      attributionControl: !noAttribution,
+    });
+
+    // Force re-render
+    setState({});
```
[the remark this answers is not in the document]
atomiomi: > woudn't empty call force re-render as well?

nope, react checks prev, next values.

> `const useForceUpdate = () => useState()[1];`

wont's work - hooks must be called in component scope.
I think it the only place where we use force render.
Would not extract it just for argument encapsulation.

packages/map/src/pin_marker/index.jsx
atomiomi: Asked @peterhass to check how we can use it. 
His answer:
> ok it’s not as simple as I thought
> since those components are used within rollup and webpack, I cannot just change the rollup config because webpack wouldn’t work

Will leave it as it is

packages/map/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "lint": "eslint -f codeframe \"./**/*.{js,jsx}\"",
+    "build": "rollup --config rollup.config.js",
+    "start": "start-storybook -p 3000 --ci",
+    "start:ssr-preview": "SSR=true start-storybook -p 3000 --ci --preview-url=/ssr-iframe",
+    "test": "NODE_ENV=test TZ='GMT' mocha \"./{,!(node_modules)/**}/*.test.{js,jsx}\"",
+    "test:tofile": "NODE_ENV=test TZ='GMT' mocha --no-colors -R spec \"./{,!(node_modules)/**}/*.test.{js,jsx}\" 2>&1 | tee test-report.log"
+  },
+  "devDependencies": {
+    "@goodhood/icons": "^1.0.0-beta.2",
+    "uni-config": "^2.1.1"
```
suprMax, line 31: do we use it?

packages/map/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "check": "npm run lint && npm run test",
+    "lint": "eslint -f codeframe \"./**/*.{js,jsx}\"",
+    "build": "rollup --config rollup.config.js",
+    "start": "start-storybook -p 3000 --ci",
+    "start:ssr-preview": "SSR=true start-storybook -p 3000 --ci --preview-url=/ssr-iframe",
+    "test": "NODE_ENV=test TZ='GMT' mocha \"./{,!(node_modules)/**}/*.test.{js,jsx}\"",
+    "test:tofile": "NODE_ENV=test TZ='GMT' mocha --no-colors -R spec \"./{,!(node_modules)/**}/*.test.{js,jsx}\" 2>&1 | tee test-report.log"
+  },
+  "devDependencies": {
+    "@goodhood/icons": "^1.0.0-beta.2",
```
suprMax, line 30: bump

packages/map/sample_data/index.jsx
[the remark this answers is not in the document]
suprMax: create a ticket for this, needs to be fixed

packages/map/src/circle/utils.jsx
```diff
@@ -0,0 +1,34 @@
...
+    'circle-stroke-width': WEIGHT_L,
+    'circle-stroke-color': COLOR_BASE,
+    'circle-color': COLOR_BASE,
+    'circle-radius': 300,
+  },
+};
+
+export const getPaint = (type, radius) => ({ ...STYLES[type], 'circle-radius': radius });
+
+export const getGeoJSON = (center) => ({
```
suprMax, line 25: small test?

packages/map/src/info_marker/index.jsx
```diff
@@ -0,0 +1,30 @@
...
+      <InfoIcon className={styles.icon} />
+    </div>
+    {children}
+  </Marker>
+);
+
+InfoMarker.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  small: PropTypes.bool.isRequired,
```
suprMax, line 27: add a default prop? validations will fail without this prop passed

packages/map/src/map/utils.jsx
```diff
@@ -0,0 +1,23 @@
+import { LngLatBounds } from 'mapbox-gl';
+
+
+export const isFilledArray = (arr) => Array.isArray(arr) && arr.length > 0;
+
+export const getStyle = (credentials) => (
+  `https://api.maptiler.com/maps/${credentials.map_id}/style.json?key=${credentials.key}`
```
suprMax, line 7: are credentials prepped for production use? map setup etc?

packages/map/src/map/utils.jsx
```diff
@@ -0,0 +1,23 @@
...
+  // Reference: https://docs.mapbox.com/mapbox-gl-js/example/zoomto-linestring/
+  const initial = new LngLatBounds(bounds[0], bounds[0]);
+  const lngLat = bounds.reduce((acc, item) => acc.extend(item), initial);
+
+  return lngLat.toArray();
+};
+
+export const mergeChildrenBounds = (bounds) => {
+  if (!isFilledArray(bounds)) return undefined;
+  return bounds.flat(1);
```
suprMax, line 22: i'm not 100% sure regarding browser support, load a polyfill or flatten by hand?

packages/map/src/marker/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+    const marker = new MapboxMarker(nodeRef.current).setLngLat(position).addTo(map);
+
+    let popup;
+    if (popupContent) {
+      popup = new Popup({ offset: popupOffset }).setDOMContent(popupRef.current);
+      marker.setPopup(popup);
+    }
+
+    if (popupDefaultState) {
```
suprMax, line 34: move into popupContent check, may be falsely triggered by omitting popupContent

packages/map/src/pin_marker/index.jsx
suprMax: create a ticket please, i think it should be doable to update both configs

packages/map/src/pin_marker/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+  PIN_MARKER_YELLOW,
+  PIN_MARKER_RED,
+  PIN_MARKER_GRAY,
+  PIN_MARKER_BLUE,
+];
+
+const images = types.reduce((acc, type) => {
+  acc[type] = require(`./images/pin-${type}.svg`);
+  return acc;
+}, {});
```
suprMax, line 30:
```suggestion
const images = types.reduce((acc, type) => ({
   ...acc, [type]: require(`./images/pin-${type}.svg`),
}), {});
```

shorter

packages/map/src/polygon/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+    <>
+      <Source geoJsonSource={getGeoJSON(area)} id={sourceId} />
+      <Layer id={layerId} type="fill" sourceId={sourceId} paint={getFillPaint(type)} onClick={onClick} />
+      <Layer type="line" sourceId={sourceId} paint={getLinePaint(type)} />
+    </>
+  );
+};
+
+Polygon.propTypes = {
+  type: getTypeProp().isRequired,
```
suprMax, line 31: i don't think utils here make sense, besides build strips this entire section out. if you use utils it's up to tree shake to determine unused utils (not sure if works)

packages/map/src/polygon/utils.jsx
```diff
@@ -0,0 +1,86 @@
...
+  POLYGON_ACTIVE,
+  POLYGON_HIGHLIGHTED,
+  POLYGON_SOLID,
+  POLYGON_THIN,
+  POLYGON_DEFAULT,
+]);
+
+export const getGeoJSON = (area) => ({
+  type: 'geojson',
+  data: {
```
suprMax, line 79: a little test?

packages/map/src/story_utils.jsx
```diff
@@ -0,0 +1,4 @@
+export const MAP_CREDENTIALS = {
```
peterhass, line 1: open for suggestions on a better filename

suprMax: can we avoid having this file in general? run DefinePlugin or something

peterhass: Those credentials are needed in almost every story. Not sure if we should simply inline it.

Since we want to move storybook config to root package, we should keep it general - therefore using DefinePlugin would add this 'define' to every subpackage's storybook

atomiomi: maybe keep simple js file in the root local.js with all secrets and add alias in webpack for it.
no need to dotenv

peterhass:
Sounds even better! We don’t need env specific settings so a simple js file
should do it. Will prepare the change

packages/map/src/pin_marker/index.jsx
peterhass: ### Part 1: Not sure if this will work at all

You are using `require` how you would in any of our webpack apps CoreFE, BizFE, etc - you are expecting file-loader / url-loader to return a public url to that asset.

Our webpack apps aren't configured to apply those loaders somewhere in node_modules.
Since we are building a library here we have to approach it differently. 

For this specific case I would use rollups svgo plugin to inline those svgs so that the app using the package doesn't need to concern itself with serving those svgs.

For now I would manually import every single one of those svgs, so that we can be sure that they are processed by rollup. I think rollup wouldn't transform this dynamic require call.

### Part 2: Import glob

Glob import behavior is difficult to control. As soon as you need named exports it gets difficult. I researched a bit and I didn't find a import-glob plugin for webpack which could be configured the same way as `rollup-plugin-glob-import`.

I suggest the following: Let's use `require.context`. It works in webpack out of the box, webpack applies all needed loaders and for rollup we can use `rollup-plugin-require-context`.

config/local.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
suprMax, line 1: remove from being commited

packages/map/README.md
```diff
@@ -15,6 +15,7 @@ npm i prop-types // v15.x.x
 npm i clsx // v2.x.x
 npm i nebenan-ui-kit // v4.x.x
```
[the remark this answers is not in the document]
suprMax: remove this section? these seem to be in dev deps

packages/map/src/circle/index.stories.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { select, withKnobs } from '@storybook/addon-knobs';
+import config from '../../../../config';
```
suprMax, line 3: @peterhass did adding that alias work?

packages/map/src/map/hooks.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return isLocked;
+};
+
+export const useMapboxComponent = (isLocked, noAttribution) => {
+  const [, setState] = useState();
+  const ref = useRef(null);
+
+  useEffect(() => {
+    ref.current = ReactMapboxGl({
```
suprMax, line 23: perhaps old map needs to be destroyed somehow? check

packages/map/src/marker/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+      marker.togglePopup();
+    }
+
+    return () => {
+      marker.remove();
+      if (popup) popup.remove();
```
suprMax, line 40: perhaps popup needs to be destroyed first?

packages/map/src/pin_marker/index.jsx
suprMax: require.context sounds right to me

packages/map/README.md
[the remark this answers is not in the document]
atomiomi: this is for installing as a package. not development

packages/map/src/circle/index.stories.jsx
peterhass: using a alias would cause eslint to complain

```
error: Unable to resolve path to module 'config' (import/no-unresolved) at src/circle/index.stories.jsx:3:20:
  1 | import React from 'react';
  2 | import { select, withKnobs } from '@storybook/addon-knobs';
> 3 | import config from 'config';
```

Didn't want to add a exception to eslint config so I kept it like it is now

packages/map/README.md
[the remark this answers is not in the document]
suprMax: npm freaks out when peer deps are not installed so we don't have to document this

PR #11, opened 2020-07-01
Icons: Build to commonjs instead esm

peterhass:
- svgr doesn't allow to specify plugins for overall babel (only for the jsx part - it is processed separately)


packages/icons/utils/naming.js
```diff
@@ -7,7 +7,7 @@ const getIconName = (fileName) => camelCase(getSourceIconName(fileName));
 const getComponentName = (fileName) => upperFirst(getIconName(fileName));
 const getLibFileName = (fileName) => snakeCase(getSourceIconName(fileName));
 const getLibSvgFileName = (fileName) => `${getLibFileName(fileName)}.svg`;
-const getLibJsFileName = (fileName) => `${getLibFileName(fileName)}.jsx`;
```
peterhass, line 10: need to change it to .js so that node can resolve it on it's own

PR #13, opened 2020-07-10
Improvements

peterhass:
https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-30301


package-templates/react/package.json
```diff
@@ -24,8 +24,8 @@
     "check": "npm run lint && npm run test",
     "lint": "eslint -f codeframe \"./**/*.{js,jsx}\"",
     "build": "rollup --config rollup.config.js",
-    "start": "start-storybook -p 3000 --ci",
-    "start:ssr-preview": "SSR=true start-storybook -p 3000 --ci --preview-url=/ssr-iframe",
+    "start": "start-storybook -p 3000 --ci --config-dir ../../.storybook",
+    "start:ssr-preview": "SSR=true start-storybook -p 3000 --ci --preview-url=/ssr-iframe --config-dir ../../.storybook",
```
peterhass, line 28: Idea: Extract to `../../scripts/start-storybook` and `../../scripts/stop-storybook`

packages/components/package.json
```diff
@@ -21,8 +21,8 @@
     "check": "npm run lint && npm run test",
     "lint": "eslint -f codeframe \"./**/*.{js,jsx}\"",
     "build": "rollup --config rollup.config.js",
-    "start": "start-storybook -p 3000 --ci",
-    "start:ssr-preview": "SSR=true start-storybook -p 3000 --ci --preview-url=/ssr-iframe",
+    "start": "start-storybook -p 3000 --ci --config-dir ../../.storybook",
+    "start:ssr-preview": "SSR=true start-storybook -p 3000 --ci --preview-url=/ssr-iframe --config-dir ../../.storybook",
```
peterhass, line 25: https://github.com/goodhood-eu/goodhood/pull/13/files#r452882427

.storybook/main.js
```diff
@@ -2,9 +2,10 @@ const path = require('path');
 const sassFunctions = require('sass-functions');
 const sass = require('sass');
 
-const ROOT_PKG_PATH = path.join(__dirname, '../../../');
+const ROOT_PKG_PATH = path.join(__dirname, '../');
```
suprMax, line 5: will we run storybook from root or from packages? ⚖️  about the same?

.storybook/ssr/src/index.jsx
```diff
@@ -23,7 +23,7 @@ const forceServerSideRender = (story) => {
 
 const api = start(render);
 
-require(`${PKG_PATH}/.storybook/preview-ssr`);
+require(`root-pkg/.storybook/preview-ssr`);
```
suprMax, line 26: do same as in core app? `@` or `@root` and `@package`?

package-templates/react/.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 const path = require('path');
```
suprMax, line 1: if we placed .eslintrc.js/babel/mocha at the root wouldn't it work without package configs?

package-templates/react/package.json
suprMax: or start from root and serve all storybooks?

package-templates/react/.eslintrc.js
peterhass: Yes it works, I removed the 'extend' part.

But still, we need those custom rules for no-extraneous-dependencies until there is a better solution for that problem. 
https://github.com/benmosher/eslint-plugin-import/issues/1174

.storybook/main.js
peterhass: Almost yes but I think it would make a difference in load times. SSR-preview is already pretty slow because it has to load all stories of ONE package but we load all stories in all packages it could get worse.

I think it's ok for subpackages to have their own `start`, `lint`, `build` scripts

.storybook/ssr/src/index.jsx
```diff
@@ -23,7 +23,7 @@ const forceServerSideRender = (story) => {
 
 const api = start(render);
 
-require(`${PKG_PATH}/.storybook/preview-ssr`);
+require(`@root/.storybook/preview-ssr`);
```
suprMax, line 26:
```suggestion
require('@root/.storybook/preview-ssr');
```

.storybook/ssr/utils.js
```diff
@@ -46,6 +46,12 @@ const getPrerenderedContent = (webpackStats, params) => {
   return app.default(params);
 };
 
+const matchWebpackTest = (file, test) => (
+  test instanceof RegExp
+    ? test.test(file)
+    : test === file
```
suprMax, line 52: string? perhaps use .includes?

babel.config.js
```diff
@@ -1,3 +1,6 @@
+const PKG_PATH = process.cwd();
```
suprMax, line 1: not `.babelrc`?

mocha.config.js
```diff
@@ -1,5 +1,7 @@
+const path = require('path');
+
 module.exports = {
-  require: '@babel/register',
+  require: path.join(__dirname, '.mocha/babel-require.js'),
```
suprMax, line 4: could we do the `.mocha/babel-require.js` right in this file and then let it require itself?

package-templates/react/.eslintrc.js
suprMax: there seem to be several options within the thread you posted, none work? or maybe disable package checking altogether for now?

package-templates/react/.mocharc.js
```diff
@@ -1 +1 @@
-module.exports = require("../../presets/mocha.config.js");
+module.exports = require("../../mocha.config.js");
```
suprMax, line 1:
```suggestion
module.exports = require('../../mocha.config.js');
```

i think this file shouldn't be needed, any reason to have it?

package-templates/react/package.json
```diff
@@ -24,8 +24,8 @@
     "check": "npm run lint && npm run test",
     "lint": "eslint -f codeframe \"./**/*.{js,jsx}\"",
     "build": "rollup --config rollup.config.js",
-    "start": "start-storybook -p 3000 --ci",
-    "start:ssr-preview": "SSR=true start-storybook -p 3000 --ci --preview-url=/ssr-iframe",
+    "start": "../../scripts/start-storybook",
+    "start:ssr-preview": "../../scripts/start-storybook --ssr-preview",
```
suprMax, line 28: shouldn't we default to SSR?

packages/components/.mocharc.js
```diff
@@ -1 +1 @@
-module.exports = require("../../presets/mocha.config.js");
```
suprMax, line 1: same as above

packages/map/.mocharc.js
```diff
@@ -1 +1 @@
-module.exports = require("../../presets/mocha.config.js");
+module.exports = require("../../mocha.config.js");
```
suprMax, line 1: same as above

mocha.config.js
peterhass: Not sure what you mean. Had to extract the whole require part to babel-require.js because I have to pass options to babel. Found no other way to do that

package-templates/react/package.json
peterhass: No, SSR is much slower.

package-templates/react/.mocharc.js
peterhass: No, you are right. Should be removed. Thought mocha doesn't have config lookup like babel does but after renaming mocha.config.js to .mocharc.js it worked

babel.config.js
peterhass: Looks like .babelrc.js doesn't get resolved.
 .babelrc -> json, .babelrc.js -> not working

.storybook/ssr/utils.js
peterhass: Not sure if this line is actually needed at all. Almost all rules use a regex `test` (at least the ones we want to target here).

A string `test` may need a more complex check because file path could be relative or something like that? Not sure? 

I suggest: Always return false for strings and add a comment

package-templates/react/.eslintrc.js
peterhass: - tried eslint-import-resolver-lerna with little success
- Didn't want to do the crazy stuff this guy did https://github.com/benmosher/eslint-plugin-import/issues/1174#issuecomment-509965883 Would rather have subpackage specific eslintrc.js for now

.storybook/ssr/utils.js
suprMax: not sure myself, leave a comment yes

PR #14, opened 2020-07-13
CircleCI and stylelint integration

peterhass:
- @suprMax needs to do CircleCI setup because he is the almighty repo admin


packages/components/src/micro_helmet/context.jsx
```diff
@@ -2,7 +2,6 @@ import { createContext } from 'react';
 
 
 export default createContext({
-  addProps() {
-    console.warn('Please use MicroHelmet Provider');
+  addProps() {    console.warn('Please use MicroHelmet Provider');
```
suprMax, line 5: roll back the test error

reports/components-eslint.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
```
suprMax, line 1: ignore reports

package-templates/react/package.json
```diff
@@ -22,12 +22,18 @@
     "reinstall": "rm -rf node_modules package-lock.json && npm install",
     "prepublishOnly": "npm run check && npm run build",
     "check": "npm run lint && npm run test",
-    "lint": "eslint -f codeframe \"./**/*.{js,jsx}\"",
+    "check:tofile": "npm run lint:tofile; npm run test:tofile",
+    "eslint": "eslint -f codeframe \"./**/*.{js,jsx}\"",
+    "eslint:tofile": "npm run eslint -- --format junit -o ../../reports/react-template-eslint.xml",
+    "stylelint": "stylelint \"./**/*.scss\" ",
+    "stylelint:tofile": "npm run stylelint -- --custom-formatter '../../node_modules/stylelint-junit-formatter' -o ../../reports/react-template-stylelint.xml",
```
peterhass, line 29: Will try to run all those check:tofile scripts directly in root package so that we don't have to copy&paste this mess into every subpackage

.circleci/config.yml
```diff
@@ -0,0 +1,27 @@
+version: 2
+jobs:
+  build:
+    branches:
+      ignore:
+        - master # only test new branches
+    docker:
+      - image: circleci/node:12
```
atomiomi, line 8: why not github actions?

peterhass: I think circleci has nicer formatting of linting/test results. 

Added a github workflow to this branch so we can try it out. It's almost the same as the circleci config but I found no quick solution on how to manage junit formatted test results

peterhass: Ok I thought it's gonna show the action as soon as I have the workflow file. Won't spend any more 'unplanned' time on this.

Do you have experience with github actions?

package-templates/react/package.json
peterhass: Stylelint works in root directory but eslint doesn't. Throws some weird internal error.

suprMax: might be related to node_modules/eslintignore configuration

peterhass: I finally moved everything into the root package. Works great, way less noise in subpackages. 

(Take a look at .eslintrc)

PR #16, opened 2020-07-13
Map beta

atomiomi:


presets/react/rollup.config.js
```diff
@@ -72,8 +72,7 @@ export default (pkg, pkgPath) => ({
       babelHelpers: 'runtime',
       rootMode: 'upward',
       exclude: [
-        path.join(pkgPath, 'node_modules/'),
-        path.join(ROOT_PKG_PATH, 'node_modules/'),
+        '**/node_modules/**/*',
```
suprMax, line 75: just `exclude: /node_modules/` ?

suprMax: ☝️

peterhass: Not sure if I tried regex. Will check

presets/react/rollup.config.js
```diff
@@ -55,7 +57,9 @@ export default (pkg, pkgPath) => ({
     peerDepsExternal(),
     postcss({
       extract: true,
-      modules: true,
+      modules: {
+        globalModulePaths: [/mapbox-gl/],
```
peterhass, line 61: Postcss plugin doesn't have `exclude` settings like other plugins do. So I think using this should be okay.

Should we process any file from node_modules with postcss? If yes, then we have to move these settings to map's own rollup config. If not, then `/node_modules/` should be ok.

suprMax: i think we have to assume same as with javascript, that node_modules includes precompiled CSS.

packages/map/package.json
```diff
@@ -11,13 +11,13 @@
...
   },
-  "version": "0.0.0",
+  "version": "1.0.0-beta.3",
   "module": "lib/index.esm.js",
   "main": "lib/index.js",
   "sideEffects": false,
   "scripts": {
     "reinstall": "rm -rf node_modules package-lock.json && npm install",
-    "prepublishOnly": "(cd ../../ && npm run check) && npm run build",
+    "prepublishOnly": "npm run build",
```
peterhass, line 20: we should keep checking before publishing

packages/map/src/pin_marker/index.jsx
```diff
@@ -24,18 +24,19 @@ const types = [
...
 
-const images = types.reduce((acc, type) => {
-  acc[type] = require(`./images/pin-${type}.svg`);
-  return acc;
-}, {});
+// TODO: move to lib on build step
+// const images = types.reduce((acc, type) => {
+//   acc[type] = require(`./images/pin-${type}.svg`);
+//   return acc;
+// }, {});
```
peterhass, line 31: remove?

atomiomi: will remove in next PR

PR #18, opened 2020-07-14
Fix linting errors

peterhass:


package.json
```diff
@@ -13,6 +13,7 @@
   },
   "private": true,
   "devDependencies": {
+    "@goodhood/icons": "^1.2.0",
```
peterhass, line 16: Had to move this dep from maps package up here. No idea why but otherwise linting would fail (ONLY ON CI https://github.com/goodhood-eu/goodhood/runs/870084327?check_suite_focus=true).

I tried so many things but I wasn't even able to locally reproduce the issue.

atomiomi: try to remove all node_modules folders in goodhood repo, then run npm i in root. 
try to build map pacakge.

PR #21, opened 2020-08-13
Remove react-mapboxgl

atomiomi:
TODO:
- Not sure if css classnames are unique enough .index-module_container__1yE_D ✓
- Would expect Circle to use meters for radius (no idea in what unit it is provided now) ✓
- Polygon is missing the pointer cursor if onClick is provided ✓
- Fix SSR preview ✓
- Update all deps to most recent version: `npx npm-check-updates -u`


peterhass:
SSR preview is dead. Looks like it tries to inject style tags into dom but there is no dom. Will need to check if storybook ssr webpack config is ok (there should be no style-loader in SSR prerender)

```
ReferenceError: document is not defined
    at insertStyleElement (main.95631c45850d8f5ad97d.bundle.js:56854:15)
    at addStyle (main.95631c45850d8f5ad97d.bundle.js:56969:13)
    at modulesToDom (main.95631c45850d8f5ad97d.bundle.js:56842:18)
    at ../../node_modules/style-loader/dist/runtime/injectStylesIntoStyleTag.js.module.exports (main.95631c45850d8f5ad97d.bundle.js:57000:25)
    at Object../node_modules/mapbox-gl/dist/mapbox-gl.css (main.95631c45850d8f5ad97d.bundle.js:57750:14)
    at __webpack_require__ (main.95631c45850d8f5ad97d.bundle.js:20:30)
    at Module../src/map/index.jsx (main.95631c45850d8f5ad97d.bundle.js:59895:86)
    at __webpack_require__ (main.95631c45850d8f5ad97d.bundle.js:20:30)
    at Module../src/circle/index.stories.jsx (main.95631c45850d8f5ad97d.bundle.js:57995:62)
    at __webpack_require__ (main.95631c45850d8f5ad97d.bundle.js:20:30)
Error: Error evaluating app script
    at getPrerenderedContent (/Users/peter/Documents/Business/nebenan.de/goodhood/.storybook/ssr/utils.js:43:11)
    at /Users/peter/Documents/Business/nebenan.de/goodhood/.storybook/ssr/middleware.js:37:27
    at Layer.handle [as handle_request] (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/layer.js:95:5)
    at /Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/index.js:281:22
    at Function.process_params (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/index.js:335:12)
    at next (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/index.js:275:10)
    at middleware (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/webpack-hot-middleware/middleware.js:38:48)
    at Layer.handle [as handle_request] (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/index.js:317:13)
    at /Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/index.js:335:12)
    at next (/Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/express/lib/router/index.js:275:10)
    at /Users/peter/Documents/Business/nebenan.de/goodhood/node_modules/webpack-dev-middleware/lib/middleware.js:40:21

```

packages/map/src/layer/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+        map.off('click', layerId, onClick);
+      }
+    };
+  }, [map, paint, onClick]);
+
+  return null;
+};
+
+Layer.propTypes = {
+  type: PropTypes.string.isRequired,
```
peterhass, line 45: provide constants for `type` prop? (calling app should not need to know about @goodhood/map's implementation details)

packages/map/src/map/hooks.jsx
```diff
@@ -1,65 +1,69 @@
...
+    if (map && boundingBox && !defaultView) {
+      map.fitBounds(boundingBox, { animate, padding: fitPadding });
+    }
+  }, [boundsSignature, map]);
 };
 
-export const useContextValue = () => {
+export const useContextValue = (map) => {
   const [bounds, setBounds] = useState([]);
+  const addBounds = (value) => {
```
peterhass, line 57: use `useCallback` to prevent unneeded renders ? (context will cause a re-render because `addBounds` is always a different object)

packages/map/src/map/utils.jsx
```diff
@@ -21,3 +21,36 @@ export const mergeChildrenBounds = (bounds) => {
   if (!isFilledArray(bounds)) return undefined;
   return bounds.flat(1);
 };
+
+export const getMapOptions = ({
```
peterhass, line 25: Maybe one test case for this whole thing? (Check if options schema is alright?) Wouldn't test every possible case.

packages/map/src/layer/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+      id: layerId,
+    });
+
+    if (onClick) map.on('click', layerId, onClick);
+
+    return () => {
+      if (map && !isDestroyed) {
+        map.removeLayer(layerId);
+        map.removeSource(layerId);
+        map.off('click', layerId, onClick);
```
suprMax, line 36: missed onClick check

do you need isDestroyed check for unsub?

packages/map/src/layer/index.jsx
suprMax: also validate with with `oneOf`?

packages/map/src/map/hooks.jsx
```diff
@@ -1,65 +1,69 @@
...
 import { getMedia, media } from '../utils';
 import MapContext from './context';
 
 
-export const useLocked = (locked, lockedMobile) => {
-  const [isLocked, setLocked] = useState(locked);
+export const useMapContext = () => useContext(MapContext);
+
+export const useMapInstance = (nodeRef, options) => {
+  const [mapInstance, setMap] = useState(false);
```
suprMax, line 12: shouldn't you use `useRef` here?

packages/map/src/map/hooks.jsx
```diff
@@ -1,65 +1,69 @@
...
 
-  return isLocked;
-};
+    const map = new Map(mapOptions);
 
-export const useMapboxComponent = (isLocked, noAttribution) => {
-  const [, setState] = useState();
-  const ref = useRef(null);
+    const handleLoad = () => {
+      if (onLoad) onLoad();
```
suprMax, line 25: `invoke`?

packages/map/src/map/hooks.jsx
```diff
@@ -1,65 +1,69 @@
...
+  const { noAttribution, locked, lockedMobile, onLoad } = options;
 
   useEffect(() => {
-    const isMobile = !getMedia(global, media.mediaM);
-    setLocked(isMobile ? lockedMobile : locked);
-  }, [locked, lockedMobile]);
+    const mapOptions = getMapOptions({
+      ...options,
+      node: nodeRef.current,
+      isMobile: !getMedia(global, media.mediaM),
```
suprMax, line 19: use hook?

packages/map/src/map/hooks.jsx
```diff
@@ -1,65 +1,69 @@
...
-  return ref.current;
+  return mapInstance;
 };
 
-export const useDefaultCenterAndZoom = (defaultZoom, defaultView) => {
-  const zoomRef = useRef(defaultZoom ? [defaultZoom] : undefined);
-  const centerRef = useRef(defaultView || undefined);
+export const useMapUpdate = (map, { bounds, childrenBounds, animate, fitPadding, defaultView }) => {
+  const boundsToFit = bounds || mergeChildrenBounds(childrenBounds);
+  const boundsSignature = JSON.stringify(boundsToFit);
```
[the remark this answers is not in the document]
suprMax: map should react to changes. i think we can optimize this call if it's slow. bounds should be a smallish array, we can do a quick jsperf test comparing this to array.join()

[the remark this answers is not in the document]
peterhass: My point wasn't that map shouldn't react to `bounds` prop changes. A `===` comparison would do fine, not shallowEqual or similar. The caller should memorize whatever gets put into `bounds` - most components work like this - why should this one behave differently?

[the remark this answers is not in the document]
atomiomi: strnigify is 54% faster

packages/map/src/circle/utils.jsx
```diff
@@ -20,7 +19,29 @@ const STYLES = {
   },
 };
 
-export const getPaint = (type, radius) => ({ ...STYLES[type], 'circle-radius': radius });
+const EARTH_CIRCUMFERENCE = 2 * Math.PI * 6378137;
```
suprMax, line 22: 6378137 -> magic number? explain

peterhass: changed it to `EQUATORIAL_RADIUS`

peterhass:
Ending notes for this issue (@atomiomi): 
- Marker Popups now work a bit differently than with our old implementation. Before, if the app rendered 10 markers with `defaultEnabled=true` one after another, only one popup was shown at a time. Now those popups work independently from one another.
- Added a performance story to check out mapbox / map package limitations. Changing circle radius is pretty fast now, changing layer source data is quite slow in comparison
   - try 'Polygon Source Change' with Circle Points = 4, Rotation Time = 2000, Angle Precision = 4: Since we render two layers for one polygon, you see them getting out of sync (at least on my machine)
- IMO: I think we may need a more flexible API for Popups. At the moment they are hard coded into Markers but it would be great to be able to show Popups on Polygon-hover for example

packages/map/src/map/hooks.jsx
atomiomi: no need. it's then used in useMemo

packages/map/src/map/hooks.jsx
atomiomi: works with state

packages/map/src/map/hooks.jsx
atomiomi: don't want to add dep only for one helper

packages/map/src/layer/hooks.jsx
```diff
@@ -0,0 +1,59 @@
...
+      map.off('click', layerId, onClick);
+      map.off('mousemove', layerId, setPointerCursor);
+      map.off('mouseleave', layerId, resetCursor);
+    };
+  }, [onClick]);
+};
+
+export const useLayerPaint = (layerId, paint) => {
+  useMapEffect((map) => {
+    if (!hasLayer(map, layerId)) return;
```
peterhass, line 56: why is this needed here?

packages/map/src/marker/index.jsx
```diff
@@ -13,37 +13,47 @@ const Marker = ({
...
-  const map = useContext(MapContext);
-  useChildrenBounds([position]);
+  const [, forceRender] = useState();
 
-  useEffect(() => {
-    if (!map) return;
+  useChildrenBounds([position]);
 
+  useMapEffect((map) => {
+    nodeRef.current = document.createElement('div');
```
peterhass, line 24: If we really need to create the element multiple times, then add a comment (see line 29)

peterhass (approved):
Looks good! Tried it in the reach prototype - no issues.

packages/map/src/layer/hooks.jsx
atomiomi: if layer is removed for example on update then mapbox throws an error because it can't find layer

packages/map/src/map/hooks.jsx
```diff
@@ -1,65 +1,101 @@
...
 import MapContext from './context';
 
 
-export const useLocked = (locked, lockedMobile) => {
-  const [isLocked, setLocked] = useState(locked);
+export const useMapContext = () => useContext(MapContext);
+
+export const useMapEffect = (fn, deps) => {
+  const { map } = useMapContext();
+  if (deps) deps.push(map);
```
suprMax, line 13: add deps=[] in arguments?

packages/map/src/map/hooks.jsx
```diff
@@ -1,65 +1,101 @@
...
+
+    const destroy = fn(map);
 
-  return isLocked;
+    let isRemoved = false;
+    const removeHandler = () => { isRemoved = true; };
+    map.on('remove', removeHandler);
+
+    return () => {
+      if (!isRemoved) {
```
suprMax, line 25: what are we solving here? multiple calls? race condition? seems like a complicated workaround

packages/map/src/map/hooks.jsx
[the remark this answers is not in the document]
suprMax: @peterhass the problem is that we calculate new array via mergeChildrenBounds, if we memo that or move the calculation inside effect perhaps we could do strict equal comparison on the array. @atomiomi ?

packages/map/src/map/utils.jsx
```diff
@@ -4,7 +4,7 @@ import { LngLatBounds } from 'mapbox-gl';
 export const isFilledArray = (arr) => Array.isArray(arr) && arr.length > 0;
 
 export const getStyle = (credentials) => (
-  `https://api.maptiler.com/maps/${credentials.map_id}/style.json?key=${credentials.key}`
+  credentials ? `https://api.maptiler.com/maps/${credentials.map_id}/style.json?key=${credentials.key}` : ''
```
suprMax, line 7: just add credentials={} in args?

packages/map/src/marker/index.jsx
```diff
@@ -13,37 +13,47 @@ const Marker = ({
...
+  if (popupRef.current) {
+    popupNode = createPortal(popupContent, popupRef.current);
+  }
+
+  let markerNode;
+  if (nodeRef.current) {
+    markerNode = createPortal(children, nodeRef.current);
+  }
+
+  return <>{popupNode}{markerNode}</>;
```
suprMax, line 56: what was the problem with just rendering these div nodes with react?

packages/map/src/map/hooks.jsx
peterhass: race condition

you have to listen to the map getting removed, otherwise it will throw as far as I know

suprMax: can we refactor this in a cleaner way or leave a comment?

peterhass: Wouldn't know how to do it cleaner

packages/map/src/layer/hooks.jsx
```diff
@@ -0,0 +1,59 @@
...
+  }, []);
+};
+
+export const useLayerSource = (layerId, geoJsonSource) => {
+  useMapEffect((map) => {
+    if (!geoJsonSource) return;
+
+    const source = map.getSource(layerId);
+    source.setData(geoJsonSource.data);
+  }, [geoJsonSource]);
```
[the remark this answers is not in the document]
atomiomi: no need. Layer id doesn't change. It's created only on first render

packages/map/src/layer/hooks.jsx
```diff
@@ -0,0 +1,59 @@
...
+      source: geoJsonSource,
+      id: layerId,
+    });
+
+    return () => {
+      map.removeLayer(layerId);
+      // Need to remove source here because it can only be removed after layer
+      map.removeSource(layerId);
+    };
+  }, []);
```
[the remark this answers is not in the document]
atomiomi: yes. repaint and source update are in separate hooks

packages/map/src/map/hooks.jsx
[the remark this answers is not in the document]
atomiomi: this can be children bounds. which we merge -> new array every call.

packages/map/src/marker/index.jsx
atomiomi: mapbox moves marker/popup nodes to body. because of that react throws error on unmount

packages/map/src/circle/utils.jsx
```diff
@@ -2,6 +2,9 @@ import { COLOR_ACTION, COLOR_BASE, WEIGHT_L, OPACITY_S } from '../constants';
 import { CIRCLE_ACTIVE, CIRCLE_DEFAULT } from './constants';
 
 
+const EARTH_CIRCUMFERENCE = 2 * Math.PI * 6378137;
+const MAPBOX_TILE_WIDTH = 512;
```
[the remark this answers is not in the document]
peterhass: Can't find any way to get that constant from mapbox.

packages/map/src/marker/index.jsx
suprMax: please add a comment

packages/map/src/layer/hooks.jsx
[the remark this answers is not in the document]
suprMax: if it doesnt change no harm in adding it, it's not clear from this file

packages/map/src/layer/hooks.jsx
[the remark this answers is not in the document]
suprMax: if it doesnt change no harm in adding it, it's not clear from this file

suprMax (approved):
LGTM

PR #22, opened 2020-08-26
Maps: Fix image packaging

peterhass:
Issues:
- Map package was missing strategies for packaging images
- Map doesn't need css-loader, css comes already pre-packaged

Issues not fixed in this PR:
- Not sure if css classnames are unique enough `.index-module_container__1yE_D`
- Would expect `Circle` to use meters for `radius` (no idea in what unit it is provided now)
- `Polygon` is missing the pointer cursor if `onClick` is provided


peterhass:
We really need to think about how to package svgs. I personally think it would make sense to pipe them through svgr so that we don't need to optimize our source svgs ourself. 

The challenge here is to configure rollup and storybook webpack the same

atomiomi (approved):

suprMax:
I would say we treat SVGs as any other image in these repositories for now, we have no need to treat them as React components. Alternatively we do treat them as React components, but then we color the marker pins programmatically instead of including every variation. For now, imo both options are acceptable

PR #23, opened 2020-08-27
Cookies Banner Component

atomiomi:


packages/components/src/cookies_banner/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+  children,
+  content,
+  GTMUrl,
+  insertGTM,
+  dismissed,
+  onScriptCreate,
+  ...rest
+}) => {
+  if (insertGTM) return <Script url={GTMUrl} onCreate={onScriptCreate} />;
+  if (dismissed) return null;
```
suprMax, line 20: this is probably not necessary/can be moved outside `condition && <Component />`

packages/components/src/cookies_banner/index.stories.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { BrowserRouter } from 'react-router-dom';
+import { text, boolean, withKnobs } from '@storybook/addon-knobs';
+
+import CookiesBanner from './index';
+
+export default { title: 'CookiesBanner', component: CookiesBanner, decorators: [withKnobs] };
+
+export const Default = () => (
+  <BrowserRouter>
```
suprMax, line 10: make this `<BrowserRouter>` wrapping a default? @peterhass thoughts?

suprMax (approved):

packages/components/src/cookies_form/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+}) => {
+  const handleValidSubmit = ({ tracking }) => {
+    onSubmit({
+      dismissed: true,
+      tracking,
+      marketing: tracking,
+      timestamp: Date.now(),
+    });
+
+    if (!tracking && defaultModel.tracking) {
```
suprMax, line 31: perhaps we should move reload also to the outside handler?

packages/components/src/cookies_banner/index.jsx
atomiomi: it's still needs to render <Script />

packages/components/src/cookies_banner/index.stories.jsx
atomiomi: @peterhass pls create a ticket in favro. Need to merge this

packages/components/src/cookies_form/index.jsx
atomiomi: will be duplicated in each fe repo

packages/components/src/cookies_banner/index.jsx
suprMax: perhaps take `Scrip` out of here? would remove 4 props including `dismissed` and peer dependency

packages/components/src/cookies_banner/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+const CookiesBanner = ({
+  className,
+  children,
+  content,
+  ...rest
+}) => (
+  <article {...rest} className={clsx(className, styles.root)}>
+    <div className={styles.wrapper}>
+      <Markdown
```
suprMax, line 16: can be a 1 liner

suprMax (approved):

PR #29, opened 2020-09-27
Reaction components

peterhass:
https://jira.nebenan.de/browse/CORE-6087


packages/components/src/reaction_selection_menu/index.module.scss
```diff
@@ -0,0 +1,86 @@
...
+$icon-size: 40px;
+
+.root {
+  touch-action: none;
+  user-select: none;
+  -webkit-touch-callout: none;
+}
+
+.listWrapper {
+  width: 100%;
```
peterhass, line 12: doesn't work? no idea why

packages/components/src/reaction_selection_menu/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+      setIsActive(true);
+    }, []),
+    onEnd: useCallback(() => {
+      setHoverReaction(null);
+      setIsActive(false);
+
+      if (hoverReaction) invoke(onSelect, hoverReaction);
+    }, [hoverReaction, onSelect]),
+    onMove: useCallback((event) => {
+      const { clientX, clientY } = event.touches[0];
```
suprMax, line 28: this is iffy across devices, you might want to use `eventCoordinates` from `nebenan-helpers/lib/dom`

packages/components/src/reaction_selection_menu/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+      const items = itemRefs.current;
+
+      if (!items.filter(Boolean).length) return;
+
+      const selectedIndex = getItemIndexForPositionFromElements(items, { x: clientX, y: clientY });
```
suprMax, line 33: naming 🔥

packages/components/src/reaction_selection_menu/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <aside
+      ref={rootRef}
+      className={clsx(styles.root, className)}
+      {...touchEvents}
+      {...mouseEvents}
+    >
+      <header>{label}</header>
+      {isActive && (
+        <article className={styles.listWrapper}>
```
suprMax, line 81: purpose of wrapper? could just render list?

packages/components/src/reaction_selection_menu/index.module.scss
```diff
@@ -0,0 +1,86 @@
...
+}
+
+.reactionLabel {
+  padding: 0 5px;
+
+  display: none;
+  position: absolute;
+  bottom: 100%;
+
+  background-color: rgb(128 128 128 / 50%);
```
suprMax, line 55: what is this magical syntax?

packages/components/src/reaction_selection_menu/utils.jsx
```diff
@@ -0,0 +1,27 @@
+import { position, screenPosition } from 'nebenan-helpers/lib/dom';
+
+const VERTICAL_THRESHOLD = 30;
+export const getItemIndexForPosition = (positions, { top, height }, { x, y }) => {
+  const isAboveElement = y < top - VERTICAL_THRESHOLD;
+  const isBelowElement = y > top + height + VERTICAL_THRESHOLD;
+  if (isAboveElement || isBelowElement) return null;
+
+  let touchedIndex = 0;
+  for (let i = positions.length - 1; i >= 0; i -= 1) {
```
suprMax, line 10: can't we get away with touchenter/touchleave on nodes?

packages/components/src/reactions_row/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+import { REACTIONS } from '../constants';
+
+const ReactionsRow = ({
+  withCounter = true,
+  reactions,
+  onClick,
+  className: passedClassName,
+  ...cleanProps
+}) => {
+  const reactionsList = useMemo(() => getSortedReactionKeys(reactions), [reactions]);
```
suprMax, line 16: same deps, merge into 1 call?

packages/components/src/constants.jsx
```diff
@@ -0,0 +1,11 @@
+export const REACTION_THANK_YOU = 'thankyou';
+export const REACTION_LOVE = 'love';
```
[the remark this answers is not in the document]
peterhass: need it in multiple components

packages/components/src/reaction_icon/index.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+// TODO: move final icons to @goodhood/icons
+const ICONS = {
+  [REACTION_THANK_YOU]: thankYouIcon,
+  [REACTION_GOOD_IDEA]: goodIdeaIcon,
+  [REACTION_BRAVO]: bravoIcon,
+  [REACTION_LOVE]: loveIcon,
+};
+
+const ReactionIcon = ({ reaction, ...cleanProps }) => (
```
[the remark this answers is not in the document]
peterhass: 1. Wanted to spare us from using svgr in the package. It's not quite easy because we run the code one time through webpack (storybook) and one time through rollup. We did the same thing in maps https://github.com/goodhood-eu/goodhood/blob/3d5942e4b60054db5790ed3e7ff36b0131b95990/packages/map/src/pin_marker/index.jsx#L9
2. We need the reactions-string-to-icon-conversion multiple times in the app. That's why I created this component

packages/components/src/reaction_selection_menu/hooks.jsx
```diff
@@ -0,0 +1,120 @@
+import { useCallback, useEffect, useRef } from 'react';
+import useMounted from 'nebenan-react-hocs/lib/use_mounted';
+import { HOVER_ENTER_TIMEOUT, HOVER_LEAVE_TIMEOUT, LONG_TOUCH_DURATION } from './constants';
+
+export const useActiveEventListener = (ref, type, listener) => {
```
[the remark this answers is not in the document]
peterhass: I think I will go with `useEventListener(ref, type, listener, options = { passive: true }`

packages/components/src/reaction_selection_menu/hooks.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+    ref.current.addEventListener(type, listener, { passive: false });
+
+    return () => {
+      ref.current.removeEventListener(type, listener);
+    };
+  }, [ref, type, listener]);
+};
+
+const preventTextSelection = (event) => event.preventDefault();
```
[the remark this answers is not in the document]
peterhass: Not really. It prevents text selection. Thought about adding a comment explaining that but I think a function call is a better fit.

packages/components/src/reaction_selection_menu/hooks.jsx
```diff
@@ -0,0 +1,120 @@
...
+  const handleTimer = useCallback((event) => {
+    if (!isMounted.current) return;
+
+    active.current = true;
+    onStart(event);
+  }, [onStart, active, isMounted]);
+
+  const handleStart = useCallback((event) => {
+    preventTextSelection(event);
+    timer.current = setTimeout(handleTimer.bind(undefined, event), LONG_TOUCH_DURATION);
```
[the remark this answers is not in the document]
peterhass: You are right, it's not actionable anymore when it reaches the handler but you can still receive values from it. 

Just looked it up. The only used event is `onMove` so we should be good to remove the other ones.

packages/components/src/reaction_selection_menu/hooks.jsx
```diff
@@ -0,0 +1,120 @@
...
+    active.current = false;
+
+    onEnd(event);
+  }, [isMounted, active, onEnd]);
+
+  const handleLeave = useCallback((event) => {
+    if (enterTimer.current) clearTimeout(enterTimer.current);
+    if (!active.current) return;
+
+    leaveTimer.current = setTimeout(handleLeaveTimer.bind(undefined, event), HOVER_LEAVE_TIMEOUT);
```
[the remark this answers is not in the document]
peterhass: Yes. Was requested by design team and it's feels weird without

packages/components/src/reaction_selection_menu/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+import { REACTIONS } from '../constants';
+
+const ReactionSelectionMenu = ({ className, label, strings, onSelect }) => {
+  const rootRef = useRef(null);
+  const [isActive, setIsActive] = useState(false);
+  const [hoverReaction, setHoverReaction] = useState(null);
+  const itemRefs = useRef([]);
+
+  const { active: activeTouchEvents, passive: touchEvents } = useLongTouch({
+    onStart: useCallback(() => {
```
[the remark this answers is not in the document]
peterhass: `useMemo` would be slightly shorter but less easy to understand.

The hook should not need to think about dependencies of the provided `onStart` callback. Inside the hook, the callback is used like usual - don't really get what you mean with double wrapping

packages/components/src/reaction_selection_menu/index.jsx
peterhass: open for suggestions

packages/components/src/reaction_selection_menu/index.jsx
peterhass: not sure why I have it or why I needed it. will take a look

packages/components/src/reaction_selection_menu/utils.jsx
peterhass: I think this was one of my first approaches but it failed. Don't remember why. Will try it when I have some time left.

packages/components/rollup.config.js
```diff
@@ -1,4 +1,13 @@
+import url from '@rollup/plugin-url';
```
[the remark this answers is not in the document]
peterhass: https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-33024

Created a favro ticket because I would need to change the map package as well -> possible merge conflicts with other PR

packages/components/src/constants.jsx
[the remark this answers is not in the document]
suprMax: move into a folder then?

packages/components/src/reaction_selection_menu/index.jsx
suprMax: `getHoverIndex` ?

packages/components/src/reaction_selection_menu/index.module.scss
```diff
@@ -0,0 +1,82 @@
...
+
+.reactionLabel {
+  margin-bottom: 5px;
+  padding: 0 5px;
+
+  display: none;
+  position: absolute;
+  bottom: 100%;
+
+  background-color: rgb(128 128 128);
```
suprMax, line 51: rgb(128, 128, 128);

packages/components/src/reaction_selection_menu/hooks.jsx
```diff
@@ -21,36 +21,34 @@ export const useLongTouch = ({ onStart, onEnd, onMove }) => {
   const active = useRef(false);
   const isMounted = useMounted();
 
-  const handleTimer = useCallback((event) => {
+  const handleTimer = useCallback(() => {
```
suprMax, line 24: do we need to wrap all these handlers in useCallback? They're not passed into useEffect etc, perf difference would be negligible no?

peterhass: I think for `handleStart`, `handleMove`, `handleEnd` and `handleCancel` we should keep the `useCallback` wrapping. Other hooks like `useEventListener` listen to changes on the callbacks like `handleStart` to know if they need to reattach it. Returning a new function all the time would cause it to reattach on every render.

The timer handlers may not need `useCallback` wrapping - will check

.storybook/main.js
```diff
@@ -88,6 +88,19 @@ module.exports = {
...
+      use: [
+        {
+          loader: '@svgr/webpack',
+          options: {
+            svgo: false,
+          }
+        },
+        'url-loader'
+      ]
+    })
```
suprMax, line 102: missing `,;` eslint works fine?

order here ok? i thought url comes first. also url-loader or file-loader? i thought we go with file since we want to re-import the same files in different places

packages/components/rollup.config.js
```diff
@@ -1,4 +1,15 @@
...
 
-export default getConfig(pkg, __dirname);
+const config = getConfig(pkg, __dirname);
+export default {
+  ...config,
+
+  plugins: [
+    ...config.plugins,
+    url(),
+    svgr({ svgo: false }),
```
suprMax, line 13: why no svgo, any issues?

packages/components/src/reaction_icon/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  [REACTION_GOOD_IDEA]: GoodIdea,
+  [REACTION_BRAVO]: Bravo,
+  [REACTION_LOVE]: Love,
+};
+
+const ReactionIcon = ({
+  className: passedClassName,
+  reaction,
+  size = REACTION_SIZE_M,
+  colorize = false,
```
suprMax, line 22: the naming is a bit awkward, more grammatically correct would be `colored` I think, which is also awkward. maybe flip it? `simple` to not have colors?

packages/components/src/reaction_icon/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  ...cleanProps
+}) => {
+  const Icon = ICONS[reaction];
+  const className = clsx(styles.root, passedClassName, styles[`is-${reaction}`], {
```
suprMax, line 26: i think the `is-XXX` classnames need to be renamed to `type-xxx`

packages/components/src/reaction_icon/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+    [styles.isColorized]: colorize,
+  });
+
+  return (
+    <Icon
+      {...cleanProps}
+      className={className}
+      style={{ height: size, width: 'auto' }}
```
suprMax, line 34: safari/firefox ok? these are square, safe to set same size imo

packages/components/src/reaction_icon/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+    />
+  );
+};
+
+ReactionIcon.propTypes = {
+  className: PropTypes.string,
+  reaction: PropTypes.oneOf(Object.keys(ICONS)),
```
suprMax, line 41: ouch, might not go through the optimization stripping of proptypes, please make sure

packages/components/src/reaction_icon/index.module.scss
```diff
@@ -0,0 +1,27 @@
+@import "nebenan-ui-kit";
+
+.isColorized {
+  &.is-thankyou {
+    > path {
```
suprMax, line 5: needs direct descendant? too fragile

packages/components/src/reaction_icon_bubble/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  colorize,
+  filled = false,
+}) => {
+  const className = clsx(passedClassName, styles.root, styles[`is-${reaction}`], {
+    [styles.isFilled]: filled,
+    [styles.isColorized]: colorize,
+  });
+
+  return (
+    <span className={className} style={{ padding: getPaddingForSize(size) }}>
```
suprMax, line 27: could you do this with %?

packages/components/src/reaction_icon_bubble/index.module.scss
```diff
@@ -0,0 +1,33 @@
+@import "nebenan-ui-kit";
```
suprMax, line 1: i think you wanted to start using `@import "~nebenan-ui-kit";`

packages/components/src/reaction_icon_bubble/index.module.scss
```diff
@@ -0,0 +1,33 @@
...
+.root {
+  display: inline-flex;
+  border-radius: 100%;
+  background-color: $color-base-04;
+}
+
+.isFilled {
+  background-color: $color-gray-01;
+
+  .icon > path {
```
suprMax, line 12: fragile selector

packages/components/src/reaction_selection_menu/hooks.jsx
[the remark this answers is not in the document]
suprMax: yea i would opt for a direct call with a comment

packages/components/src/reaction_selection_menu/index.jsx
```diff
@@ -0,0 +1,101 @@
...
+      setIsActive(true);
+    }, []),
+    onEnd: useCallback(() => {
+      setHoverReaction(null);
+      setIsActive(false);
+
+      if (hoverReaction) invoke(onSelect, hoverReaction);
```
suprMax, line 27: you just set it to `null` on L24, probably need to copy it somewhere?

packages/components/src/reaction_selection_menu/index.module.scss
```diff
@@ -0,0 +1,74 @@
...
+
+.reactionLabel {
+  margin-bottom: 2px;
+  padding: 2px 8px;
+
+  display: none;
+  position: absolute;
+  bottom: 100%;
+
+  background-color: rgba(97, 116, 110, .8);
```
suprMax, line 47: does it have to be this exact shade of gray? can't we just use ui-kit colors?

packages/components/src/reaction_selection_menu/index.module.scss
```diff
@@ -0,0 +1,74 @@
...
+  color: #fff;
+  border-radius: 11.5px;
```
suprMax, line 49: i wouldn't want the rounding left to the browser

packages/components/src/reaction_selection_menu/utils.jsx
suprMax: worth investigating imo, this is not ideal

packages/components/src/reactions_row/index.jsx
```diff
@@ -0,0 +1,53 @@
...
+  ...cleanProps
+}) => {
+  const { list: reactionsList, count } = useReactionData(reactions);
+
+  if (withCounter && count === 0) return null;
+
+  const renderReaction = (reaction) => (
+    <ReactionIconBubble
+      size={REACTION_SIZE_S}
+      colorize
```
suprMax, line 23: looking through - colorized should be the default. go for `simple` to disable those styles (flip the condition in clsx: `isColored: !simple`)

packages/components/src/reactions_row/index.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <span {...cleanProps} className={clsx(passedClassName, styles.root, { 'ui-link': onClick })} onClick={onClick}>
+      {reactionsList.map(renderReaction)} {countNode}
+    </span>
+  );
+};
+
+ReactionsRow.propTypes = {
+  reactions: PropTypes.shape(REACTIONS.reduce((acc, reaction) => ({
```
suprMax, line 44: logic here might prevent optimizer from removing the validations in production build

.storybook/main.js
peterhass: it may not process hidden directories, will check

I think the order is alright - it works at least(?)

Doesn't really matter if we use file-loader or url-loader. This config only applies to storybook webpack and both loaders produce a URI

packages/components/rollup.config.js
peterhass: Didn't want to add logic for package specific svgo/svgr configs here. That's why I went with the most minimal setup for now.

peterhass: https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-33024

packages/components/src/reaction_icon/index.jsx
peterhass: I don't think using colors by default is a good choice. I would expect something called `*Icon` to be unicolor by default (like all our icons in @goodhood/icons that are named that way)

`colored` may be better

packages/components/src/reaction_icon/index.jsx
peterhass: Current icons aren't square, that's why I used auto on width.

Safari / FF look alright

packages/components/src/reaction_icon_bubble/index.jsx
peterhass: Unfortunately no, values are too different

packages/components/src/reaction_icon_bubble/index.module.scss
peterhass: Yes, I started out with ~ but I think I had issues with rollup. It wasn't able to find the file. Didn't investigate any deeper though

.storybook/main.js
suprMax: ok here, might be not ok in rollup, might need file loader there

peterhass: Would prevent using like file-loader for rollup as long as possible. Will make it more difficult to setup apps to use our packages since they would need to serve the package's static files too.

packages/components/src/reaction_icon/index.jsx
peterhass: checked it - the babel plugins removes it

packages/components/src/reaction_selection_menu/index.jsx
peterhass: Good point. This whole callback only worked because of async state updates. Will move this line up to the top

.storybook/main.js
suprMax: url loader would blow up the bundle size a lot. let's see if we need it but there might be no other way

packages/components/src/reaction_icon/index.jsx
suprMax: just call it `Reaction` and problem solved

packages/components/src/reaction_icon_bubble/index.module.scss
suprMax: worth looking into, add a ticket please

peterhass: https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-33170

packages/components/src/reactions_row/index.jsx
```diff
@@ -8,12 +8,14 @@ import ReactionIconBubble from '../reaction_icon_bubble';
...
   withCounter = true,
+  limit = REACTIONS.length,
   reactions,
   onClick,
   className: passedClassName,
   ...cleanProps
 }) => {
-  const { list: reactionsList, count } = useReactionData(reactions);
+  const { list: sortedReactionTypes, count } = useReactionData(reactions);
+  const reactionTypes = sortedReactionTypes.slice(limit * -1);
```
suprMax, line 18: `-limit`?

packages/components/src/reaction_selection_menu/utils.jsx
peterhass: `touchenter` and `touchleave` events don't exist. we only have `touchstart`, `touchend`, `touchcancel` and `touchmove`

As far as I understand, `touchmove` only gets triggered on the element that received `touchstart`. That means that I cannot use `touchmove` on reaction list items because the initial touch started on a parent

peterhass: Played around with custom events and `elementFromPoint` but it is even more complicated than the current approach. Would leave it as it is.

suprMax (approved):

PR #36, opened 2020-10-07
Added data-track for accept-button

ozerovs:


packages/components/src/cookies_form/index.jsx
```diff
@@ -33,6 +33,7 @@ const CookiesForm = ({
   if (compactView) {
     alternativeAction = (
       <span
+        data-track="cookie_form-button_primary"
```
peterhass, line 36: https://github.com/goodhood-eu/goodhood/pull/35/files

we changed the same thing

PR #43, opened 2020-10-23
Tooltips popper

ozerovs:
https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-30098


package.json
```diff
@@ -90,6 +90,7 @@
     "lint": "npm run eslint && npm run stylelint",
     "eslint": "eslint -f codeframe \"./**/*.{js,jsx}\" ",
     "stylelint": "stylelint \"./**/*.scss\" ",
-    "test": "scripts/run-tests"
+    "test": "scripts/run-tests",
+    "start": "cd packages/components && npm run start"
```
peterhass, line 94: let's not use a `start` script in the root package, every subpackage should do that on it's own

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const { type, text, children, ...cleanProps } = props;
+  const selectedType = knownTypes[type] ? type : 'top';
+  const className = clsx(styles.tooltip, props.className, {
+    [styles.isActive]: isOpen,
+  });
+
+  const handleOpen = useCallback((event) => {
+    event.stopPropagation();
+    setOpen(true);
+  }, [isOpen]);
```
peterhass, line 23: no need to use `isOpen` as a dependency since you are not using it. setters also don't need to be listed in dependencies because they are guaranteed to always be the same function

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,68 @@
+import React, { useCallback, useRef, useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import clsx from 'clsx';
+import { createPopper } from '@popperjs/core';
+
+import { arrayToHash } from 'nebenan-helpers/lib/data';
+import { getPopperOptions } from './utils';
+import styles from './index.module.scss';
+
+const knownTypes = arrayToHash(['left', 'right', 'top', 'bottom']);
```
peterhass, line 10: `KNOWN_TYPES`

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+  const handleClose = useCallback((event) => {
+    event.stopPropagation();
+    setOpen(false);
+  }, [isOpen]);
+
+  const refElement = useRef(null);
+  const refTooltip = useRef(null);
+  const refArrow = useRef(null);
+
+  useEffect(() => {
```
peterhass, line 34: I think `useLayoutEffect` is needed here, please check the documentation

https://github.com/popperjs/react-popper/blob/master/src/usePopper.js

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+    if (refElement.current && refTooltip.current) {
+      createPopper(
```
peterhass, line 36: provide a proper teardown function and add `selectedType` to dependencies list

packages/components/src/tooltip/index.module.scss
```diff
@@ -0,0 +1,75 @@
...
+    width: $tooltip-arrow-size;
+    height: $tooltip-arrow-size;
+
+    content: '';
+    position: absolute;
+
+    border-radius: $border-radius-s;
+
+    background: $color-gray-00;
+    //background: red;
```
peterhass, line 51: remove comment

packages/components/src/tooltip/index.stories.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { withKnobs } from '@storybook/addon-knobs';
+import Tooltip from './index';
+import styles from './index.stories.module.scss';
+
+const tooltipShort = 'Design is like a joke.';
+
+export default { title: 'Tooltip', component: Tooltip, decorators: [withKnobs] };
+
+export const Default = () => (
```
peterhass, line 10: use knobs to set the tooltip text and type

packages/components/src/tooltip/index.stories.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { withKnobs } from '@storybook/addon-knobs';
+import Tooltip from './index';
+import styles from './index.stories.module.scss';
+
+const tooltipShort = 'Design is like a joke.';
+
+export default { title: 'Tooltip', component: Tooltip, decorators: [withKnobs] };
```
peterhass, line 8: please try if it still works if you change `text` or `children` after initial render

peterhass: and try running it with ssr storybook

packages/components/src/feature_alert/index.stories.jsx
```diff
@@ -0,0 +1,65 @@
...
+      {`Tooltip position: ${select('Position', POSITIONING_OPTIONS, POSITION_TOP)}`}
+    </Tooltip>
+  </div>
+);
+
+export const AppearanceExamples = () => (
+  <div className={styles.container}>
+    <div className={styles.element}>
+      <Tooltip content={text('Tooltip text', TOOLTIP_PLACEHOLDER)}>
+        NO TRIGGER
```
peterhass, line 41: what's the purpose of a no-trigger tooltip? :D

packages/components/src/feature_alert/index.stories.jsx
```diff
@@ -0,0 +1,65 @@
...
+      </Tooltip>
+    </div>
+    <div className={styles.element}>
+      <Tooltip content={text('Tooltip text', TOOLTIP_PLACEHOLDER)} defaultOpen closeIcon>
+        defaultOpen
+      </Tooltip>
+    </div>
+    <div className={styles.element}>
+      <Tooltip content={text('Tooltip text', TOOLTIP_PLACEHOLDER)} trigger={TRIGGER_HOVER}>
```
peterhass, line 50: Suggestion: use the `select` knob to choose the trigger size, that way you only need one tooltip on the page - may be less overwhelming (if you need to force re-render then use `key` prop)

packages/components/src/feature_alert/hooks.jsx
```diff
@@ -0,0 +1,15 @@
+import { useEffect } from 'react';
+import keymanager from 'nebenan-helpers/lib/keymanager';
+import { invoke } from 'nebenan-helpers/lib/utils';
+
+export const useEscHandler = (callback) => useEffect(() => keymanager('esc', callback), [callback]);
+
+export const useOutsideClick = (ref, callback) => useEffect(() => {
```
peterhass, line 7: @suprMax move to hocs package?

packages/components/src/feature_alert/index.jsx
```diff
@@ -0,0 +1,2 @@
+export { default as Tooltip } from './tooltip';
```
peterhass, line 1: why don't you put the component in here?

packages/components/src/feature_alert/index.module.scss
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  100% {
+    opacity: 1;
+  }
+}
+
+.container {
+  width: 300px;
+  display: none;
```
peterhass, line 17: .isActive isn't needed. We only render the tooltip if it is active, so it's always shown as 'block'

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+    <article {...cleanProps} className={className} ref={wrapper} onClick={handleClose}>
+      { isOpen && (
+        <aside
+          className={styles.container} ref={setTooltipElement}
+          style={popperStyles.popper} {...attributes.popper}
+        >
+          <i className={styles.arrow} ref={setArrowElement} style={popperStyles.arrow} />
+          <div className={styles.content}>
+            {content}
+            {closeIcon && <i className={`${styles.cross} icon-cross`} />}
```
peterhass, line 108: use `@goodhood/icons` cross?

packages/components/src/feature_alert/index.module.scss
```diff
@@ -0,0 +1,100 @@
...
+  color: $color-blue-01;
+  background: $color-gray-09;
+  border-radius: 50%;
+}
+
+.tooltip {
+  display: inline-block;
+  position: relative;
+
+  &.isActive .container {
```
peterhass, line 96: no need for isActive anymore

packages/components/src/feature_alert/index.stories.jsx
peterhass: + one story with no custom props at all - only default props (no knobs needed)

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    return () => clearTimeout(tid);
+  }, [trigger, handleOpen]);
+
+  const className = clsx(styles.tooltip, props.className, {
+    [styles.isActive]: isOpen,
+  });
+
+  const triggerProps = !isOpen && getTriggerProps(trigger, handleOpen);
```
peterhass, line 96: Don't think you need the `isOpen` check in here. You already have it in `handleOpen`

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+  useOutsideClick(wrapper, handleClose);
+
+  useEffect(() => {
+    let tid;
+    if (trigger === TRIGGER_DELAYED && !wasActive.current) {
+      tid = setTimeout(handleOpen, DELAY_TIMEOUT);
+    }
+
+    return () => clearTimeout(tid);
+  }, [trigger, handleOpen]);
```
peterhass, line 90: - rather do a guard condition, then you can use `const` instead of `let`
- extract to hooks.jsx 'useDelayedOpen(trigger, handleOpen, timeout)`

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const className = clsx(styles.tooltip, props.className, {
+    [styles.isActive]: isOpen,
+  });
+
+  const triggerProps = !isOpen && getTriggerProps(trigger, handleOpen);
+
+  return (
+    <article {...cleanProps} className={className} ref={wrapper} onClick={handleClose}>
+      { isOpen && (
```
peterhass, line 100: styleguide: no spaces at the beginning and end of jsx execution thingy `{isOpen && <h1>hi</h1>}`

packages/components/src/feature_alert/tooltip.jsx
peterhass, line 108: should it do anything on click?

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+    children,
+    closeIcon,
+    defaultOpen,
+    onOpen,
+    onClose,
+    ...cleanProps
+  } = props;
+
+  const [isOpen, setOpen] = useState(defaultOpen);
+  const wrapper = useRef(null);
```
peterhass, line 39: `rootRef` ?

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+  // Workaround: Fix tooltip positioning when custom sizes are applied in css
+  // Positioning only breaks if container is rendered in the same cycle as
+  // `usePopper` was called for the first time.
+  useEffect(() => {
+    invoke(forceUpdate);
+  }, [forceUpdate]);
+
+  // need to be able to only open once
+  const wasActive = useRef(false);
```
peterhass, line 64: extract hook `const { isOpen, isOpenRef, setIsOpen } = useOpenState(defaultOpen)`

peterhass, line 64: - `wasActive` -> `wasActiveOnce`  ?

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const handleOpen = useCallback((event) => {
+    if (event) event.stopPropagation();
+    if (isOpen || wasActive.current) return;
+    wasActive.current = true;
+    setOpen(true);
+    invoke(onOpen);
+  }, []);
```
peterhass, line 72: missing `isOpen` in deps

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const handleClose = useCallback(() => {
+    if (!isOpen) return;
+    setOpen(false);
+    invoke(onClose);
+  }, []);
```
peterhass, line 78: missing `isOpen` in deps

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,151 @@
...
+    let tid;
+    if (trigger === TRIGGER_DELAYED && !wasActive.current) {
+      tid = setTimeout(handleOpen, DELAY_TIMEOUT);
+    }
+
+    return () => clearTimeout(tid);
+  }, [trigger, handleOpen]);
+
+  const className = clsx(styles.tooltip, props.className, {
+    [styles.isActive]: isOpen,
```
peterhass, line 93: decide for a naming scheme: rather use `isActive` everywhere or use `isOpen` everywhere

packages/components/src/feature_alert_label/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+import {
+  POSITION_TOP,
+  POSITION_BOTTOM,
+  POSITION_LEFT,
+  POSITION_RIGHT,
+} from '../feature_alert';
+import styles from './index.module.scss';
+
+
+const FeatureAlertLabel = ({ position, label, children, ...props }) => {
```
peterhass, line 13: `props` -> `cleanProps`

packages/components/src/feature_alert_label/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+  const className = clsx(styles.label, styles[`is-position-${position}`], props.className);
```
peterhass, line 14: why don't you get `className` right in fn args?

packages/components/src/feature_alert_label/index.module.scss
```diff
@@ -0,0 +1,40 @@
...
+
+  &.is-position-top .badge {
+    margin-bottom: 5px;
+    order: -1;
+  }
+}
+
+.badge {
+  color: $color-gray-09;
+  background: linear-gradient(110deg, #33a6b4, $color-blue-01);
```
peterhass, line 37: don't we have #33a6b4 in ui-kit?

packages/components/src/feature_alert_label/index.module.scss
```diff
@@ -0,0 +1,40 @@
...
+  font-size: 12px;
```
peterhass, line 38: $font-size-s?

packages/components/src/feature_alert_label/index.module.scss
```diff
@@ -0,0 +1,40 @@
...
+  font-weight: 600;
```
peterhass, line 39: I think this could be $font-weight-l - not sure though

packages/components/src/feature_alert_label/index.stories.jsx
```diff
@@ -0,0 +1,28 @@
...
+  POSITION_BOTTOM,
+  POSITION_LEFT,
+  POSITION_RIGHT,
+} from '../feature_alert';
+
+const TOOLTIP_PLACEHOLDER = 'Design is like a joke.';
+
+export default { title: 'Feature Alert Label', component: Label, decorators: [withKnobs] };
+
+const options = {
```
peterhass, line 15: `OPTIONS`

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+import styles from './index.module.scss';
+
+const KNOWN_TYPES = arrayToHash([POSITION_LEFT, POSITION_RIGHT, POSITION_TOP, POSITION_BOTTOM]);
+
+const Tooltip = (props) => {
+  const [isOpen, setOpen] = useState(false);
+  const { type, text, children, ...cleanProps } = props;
+  const selectedType = KNOWN_TYPES[type] ? type : POSITION_TOP;
+  const className = clsx(styles.tooltip, props.className, {
+    [styles.isActive]: isOpen,
```
peterhass, line 23: not needed anymore, same as in FeatureAlert tooltip

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+  POSITION_RIGHT,
+} from '../feature_alert';
+import styles from './index.module.scss';
+
+const KNOWN_TYPES = arrayToHash([POSITION_LEFT, POSITION_RIGHT, POSITION_TOP, POSITION_BOTTOM]);
+
+const Tooltip = (props) => {
+  const [isOpen, setOpen] = useState(false);
+  const { type, text, children, ...cleanProps } = props;
+  const selectedType = KNOWN_TYPES[type] ? type : POSITION_TOP;
```
peterhass, line 21: Would rather set POSITION_TOP as defaultProp and expect users to pass a working type

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+  }, []);
+
+  const [refElement, setRefElement] = useState(null);
+  const [tooltipElement, setTooltipElement] = useState(null);
+  const [arrowElement, setArrowElement] = useState(null);
+
+  const popperOptions = useMemo(
+    () => getPopperOptions(arrowElement, selectedType),
+    [arrowElement, selectedType],
+  );
```
peterhass, line 43: `usePopperOptions(arrowElement, selectedType)` -> same in FeatureAlert

packages/components/src/tooltip/index.module.scss
```diff
@@ -0,0 +1,77 @@
...
+    display: none;
+    position: absolute;
+    z-index: 1;
+
+    background: $color-gray-00;
+    border-radius: $border-radius-m;
+    box-shadow: $shadow-bottom-default;
+
+    font-size: 12px;
+    line-height: 16px;
```
peterhass, line 22: can we use variables from ui-kit?

packages/components/src/tooltip/index.module.scss
```diff
@@ -0,0 +1,77 @@
+@import 'nebenan-ui-kit';
+$tooltip-arrow-size: 5px;
```
peterhass, line 2: remove

packages/components/src/feature_alert/index.stories.jsx
ozerovs: We had that as an example in `nebenan-components` so I was thinking to leave it this way :D 
Since it's tooltip anyway, but...with no trigger.

peterhass: Shame on me for approving that PR

packages/components/src/feature_alert/index.jsx
ozerovs: 1) Couldn't figure out how to make 2 components work within one story
2) Wasn't thinking about point 1 since thought it'll be ok to move it, since we are moving away `Tooltip` component, and this case looks pretty similar to me

peterhass: ok

packages/components/src/feature_alert_label/index.module.scss
ozerovs: Nope, didn't find any

suprMax: choose nearest color?

packages/components/src/tooltip/index.module.scss
ozerovs: For line-height - doesn't seem so

packages/components/src/feature_alert_label/index.module.scss
ozerovs: The nearest one is `$color-blue-01`
So we can make it ` background: linear-gradient(110deg, $color-blue-01,  $color-blue-00);` instead, which will make it a bit darker

packages/components/src/feature_alert/hooks.jsx
```diff
@@ -0,0 +1,43 @@
...
+    if (ref.current.contains(event.target)) invoke(callback, event);
+  };
+
+  document.addEventListener('click', handler);
+
+  return () => document.removeEventListener('click', handler);
+}, [callback]);
+
+export const useDelayedOpen = (trigger, wasActiveOnce, callback) => useEffect(() => {
+  const tid = (trigger === TRIGGER_DELAYED && !wasActiveOnce.current)
```
peterhass, line 34: `if (trigger !== TRIGGER_DELAYED || wasActiveOnce.current) return;` ?

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,145 @@
...
+  );
+
+  // Workaround: Fix tooltip positioning when custom sizes are applied in css
+  // Positioning only breaks if container is rendered in the same cycle as
+  // `usePopper` was called for the first time.
+  useEffect(() => {
+    invoke(forceUpdate);
+  }, [forceUpdate]);
+
+  // need to be able to only open once
```
peterhass, line 66: remove comment - comments easily get outdated very fast

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,145 @@
...
+
+  const handleClose = useCallback(() => {
+    if (!isOpen) return;
+    setOpen(false);
+    invoke(onClose);
+  }, [isOpen]);
+
+  useEscHandler(handleClose);
+  useOutsideClick(rootRef, handleClose, closeIcon);
+  useCloseElemClick(closeIconRef, handleClose);
```
peterhass, line 85: One thing I don't really understand is what logic you are trying to achieve here. Can't we just put onClick on the tooltip container and hide onClick?

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+import { usePopper } from 'react-popper';
+
+import { usePopperOptions } from './hooks';
+import { POSITION_TOP } from '../feature_alert';
+import styles from './index.module.scss';
+
+const Tooltip = (props) => {
+  const [isOpen, setOpen] = useState(false);
+  const { type, text, children, ...cleanProps } = props;
+  const className = clsx(styles.tooltip, props.className);
```
peterhass, line 13: `props.className` -> `className` ?

packages/components/src/feature_alert/hooks.jsx
```diff
@@ -0,0 +1,43 @@
+import { useEffect, useMemo } from 'react';
+import keymanager from 'nebenan-helpers/lib/keymanager';
+import { invoke } from 'nebenan-helpers/lib/utils';
+import { TRIGGER_DELAYED, DELAY_TIMEOUT } from './constants';
+import { getPopperOptions } from './utils';
+
+export const useEscHandler = (callback) => useEffect(() => keymanager('esc', callback), [callback]);
+
+export const useOutsideClick = (ref, callback, shouldBeIgnored) => useEffect(() => {
```
peterhass, line 9: `shouldBeIgnored` -> `disabled` ?

packages/components/src/feature_alert/tooltip.jsx
ozerovs: Actually yes, seems like a bit more elegant, but seems like ended up being unnecessary sophisticated :)

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,141 @@
...
+    POSITION_BOTTOM,
+    POSITION_LEFT,
+    POSITION_RIGHT,
+  ]),
+  fallbackPosition: PropTypes.oneOf([
+    POSITION_TOP,
+    POSITION_BOTTOM,
+    POSITION_LEFT,
+    POSITION_RIGHT,
+  ]),
```
peterhass, line 127: Talked to Max, no need for fallbackPosition

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,141 @@
...
+  const [tooltipElement, setTooltipElement] = useState(null);
+  const [arrowElement, setArrowElement] = useState(null);
+
+  const popperOptions = usePopperOptions(arrowElement, position, fallbackPosition);
+
+  const { styles: popperStyles, attributes, forceUpdate } = usePopper(
+    refElement,
+    tooltipElement,
+    popperOptions,
+  );
```
peterhass, line 55: Discussed with Max, let's implement the wrapper ourself. Shouldn't be too hard

peterhass:
```
import React, { useRef, useEffect } from 'react';
import { createPopper } from '@popperjs/core';
import styles from './test_tooltip.module.scss';
const Component = ({ text }) => {
  const button = useRef(null);
  const tooltip = useRef(null);
  const instance = useRef(null);
  const openPopper = () => {
    console.warn('create');
    tooltip.current.setAttribute('data-show', '');
    instance.current = createPopper(button.current, tooltip.current, {
      modifiers: [
        {
          name: 'offset',
          options: {
            offset: [0, 8],
          },
        },
      ],
    });
  };
  const destroyPopper = () => {
    tooltip.current.removeAttribute('data-show');
    if (instance.current) instance.current.destroy();
    console.warn('destroy');
  };
  useEffect(() => {
    openPopper();
    return () => {
      destroyPopper();
    };
  }, [text]);
  return (
    <div>
      <span ref={button} className="ui-button ui-button-primary">DONT CLICK ME</span>
      <span ref={tooltip} className={styles.tooltip}>{text}</span>
    </div>
  );
};
export default Component;
```

packages/components/src/feature_alert/tooltip.jsx
ozerovs: But what's the point then for implementing Popper? :)

peterhass: I only meant that the calling component doesn't need to know about fallback positions. It should be able to set a preferred position and if that's not available then put the tooltip anywhere where there is space

packages/components/src/tooltip/index.jsx
```diff
@@ -1,65 +1,52 @@
...
-
-  const { styles: popperStyles, attributes } = usePopper(refElement, tooltipElement, popperOptions);
+const Tooltip = ({ type, text, children, className }) => {
+  const content = useRef(null);
+  const tooltip = useRef(null);
+  const instance = useRef(null);
+  const arrow = useRef(null);
+
+  const openPopper = () => {
+    tooltip.current.setAttribute('data-show', '');
```
peterhass, line 14: I don't think we need `data-show` here. The tutorial only uses it to show/hide the tooltip using css

packages/components/src/tooltip/index.jsx
```diff
@@ -1,65 +1,52 @@
...
-  const popperOptions = usePopperOptions(arrowElement, type);
-
-  const { styles: popperStyles, attributes } = usePopper(refElement, tooltipElement, popperOptions);
+const Tooltip = ({ type, text, children, className }) => {
+  const content = useRef(null);
+  const tooltip = useRef(null);
+  const instance = useRef(null);
+  const arrow = useRef(null);
+
+  const openPopper = () => {
```
peterhass, line 13: Wrap in `useCallback` if you want to keep it outside of `useEffect`, otherwise it's easy to miss `dependencies`

packages/components/src/tooltip/index.module.scss
```diff
@@ -2,27 +2,29 @@
...
 $tooltip-arrow-size: 10px;
 
+.wrap:hover .tooltip {
+  visibility: visible;
+}
+
 .tooltip {
   position: relative;
   cursor: default;
+  visibility: hidden;
```
peterhass, line 12: does positioning work with `visibility: hidden`? Didn't popper go crazy if it got initialized with `display: none`?

packages/components/src/tooltip/index.jsx
```diff
@@ -1,65 +1,52 @@
...
+    tooltip.current.setAttribute('data-show', '');
+    instance.current = createPopper(
+      content.current,
+      tooltip.current,
+      getPopperOptions(arrow.current, type),
+    );
+  };
+
+  const destroyPopper = () => {
```
peterhass, line 22: same thing here about `useCallback`

packages/components/package.json
```diff
@@ -47,8 +47,10 @@
     "string-validate": "^3.5.0"
   },
   "dependencies": {
+    "@popperjs/core": "^2.5.4",
     "escape-html": "^1.0.3",
-    "marked": "^1.2.5"
+    "marked": "^1.2.5",
+    "react-popper": "^2.2.4"
```
suprMax, line 53: still needed?

packages/components/src/feature_alert/index.stories.jsx
```diff
@@ -0,0 +1,52 @@
...
+      content={text('Tooltip text', TOOLTIP_PLACEHOLDER)}
+      position={select('Position', POSITIONING_OPTIONS, POSITION_TOP)}
+      fallbackPosition={select('Fallback position', POSITIONING_OPTIONS, POSITION_TOP)}
+      trigger={select('Trigger', TRIGGER_OPTIONS, TRIGGER_HOVER)}
+      closeIcon={boolean('Has close icon', true)}
+      defaultOpen={boolean('Is default open', false)}
+      onOpen={action('opened')}
+      onClose={action('closed')}
+
+      key={boolean('Reload tooltip', false)}
```
peterhass, line 47: any way we could make this a button?

naming suggestion: `Remount tooltip`

packages/components/src/feature_alert/hooks.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+  return () => document.removeEventListener('click', handler);
+}, [callback]);
+
+export const useDelayedOpen = (trigger, wasActiveOnce, callback) => useEffect(() => {
+  if (trigger !== TRIGGER_DELAYED || wasActiveOnce.current) return;
+
+  const tid = setTimeout(callback, DELAY_TIMEOUT);
+
+  return () => clearTimeout(tid);
```
peterhass, line 25:
```suggestion
  return () => { clearTimeout(tid); };
```

packages/components/src/feature_alert/hooks.jsx
```diff
@@ -0,0 +1,26 @@
...
+  };
+  document.addEventListener('click', handler);
+
+  return () => document.removeEventListener('click', handler);
+}, [callback]);
+
+export const useDelayedOpen = (trigger, wasActiveOnce, callback) => useEffect(() => {
+  if (trigger !== TRIGGER_DELAYED || wasActiveOnce.current) return;
+
+  const tid = setTimeout(callback, DELAY_TIMEOUT);
```
peterhass, line 23: call it `timer`?

packages/components/src/feature_alert/index.module.scss
```diff
@@ -0,0 +1,92 @@
+@import 'nebenan-ui-kit';
+
+$feature-alert-arrow-size: 16px;
```
peterhass, line 3: variable not needed imo

packages/components/src/feature_alert/index.module.scss
```diff
@@ -0,0 +1,92 @@
...
+  width: 300px;
+  position: absolute;
+  cursor: pointer;
+  z-index: 1;
+
+  animation: 300ms ease-in feature-alert-appear;
+  visibility: hidden;
+
+  &.isActive {
+    visibility: visible;
```
peterhass, line 25: Please check if using visibility could lead to other implications on the page. I think it should be fine since the tooltip is absolute positioned but better to check

packages/components/src/feature_alert/index.module.scss
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+// Application of css classes that are not being applied by popper.js
```
peterhass, line 29: comment not needed imo

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+  const rootClassName = clsx(styles.root, className);
+  const tooltipClassName = clsx(styles.tooltip, { [styles.isActive]: isOpen });
+
+  const triggerProps = getTriggerProps(trigger, handleOpen);
+
+  const bubble = (
+    <aside className={styles.content}>
+      {content}
+      {closeIcon && <span onClick={handleClose} className={styles.cross}><CrossIcon /></span>}
```
peterhass, line 74: could we pass `onClick` and `className` right to `CrossIcon` ? (-> get rid of the span)

packages/components/src/feature_alert/tooltip.jsx
peterhass: still valid, better to keep the public api slim

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+  const wasActiveOnce = useRef(false);
+
+  const handleOpen = useCallback((event) => {
+    if (event) event.stopPropagation();
+    if (isOpen || wasActiveOnce.current) return;
+    wasActiveOnce.current = true;
+    setOpen(true);
+    invoke(onOpen);
+  }, [isOpen]);
```
peterhass, line 54: `onOpen` missing in deps

packages/components/src/feature_alert/tooltip.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+  const handleClose = useCallback(() => {
+    if (!isOpen) return;
+    setOpen(false);
+    invoke(onClose);
+  }, [isOpen]);
```
peterhass, line 60: `onClose` missing in deps

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+      <span ref={tooltip} className={tooltipClassName || styles.tooltip}>
+        {bubble}
+        <i ref={arrow} className={arrowClassName || styles.arrow} />
+      </span>
+    </div>
+  );
+};
+
+Tooltip.propTypes = {
+  type: PropTypes.string.isRequired,
```
peterhass, line 59: would rather use `position` because `type` is very generic

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+    tooltip.current.removeAttribute('data-show');
+    if (instance.current) instance.current.destroy();
+  };
+
+  useEffect(() => {
+    mountPopper();
+    return () => {
+      destroyPopper();
+    };
+  }, [children]);
```
peterhass, line 43: brave move - did you test it with a value that changes every second? is it even needed?

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,75 @@
+import React, { useRef, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { createPopper } from '@popperjs/core';
+import { getPopperOptions } from './utils';
+import styles from './index.module.scss';
+
+const Tooltip = (props) => {
```
peterhass, line 7: contradicts single responsibility - we need to extract something generic

peterhass (changes requested):

packages/components/src/feature_alert/tooltip.jsx
ozerovs: There was an issue styling a CrossIcon without this particular span - that's the reason it's applied.

packages/components/src/feature_alert/index.module.scss
ozerovs: It's actually has problems if using `display` prop - it breaks the tooltip positioning

packages/components/src/feature_alert/index.module.scss
ozerovs: We use it 2 times, and it explains the reason behind 16px somehow.

packages/components/src/base_tooltip/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+  const destroyPopper = useCallback(() => {
+    if (instance.current) instance.current.destroy();
+  }, [instance]);
+
+  useEffect(() => {
+    mountPopper();
+    return () => {
+      destroyPopper();
+    };
+  }, [children]);
```
peterhass, line 39: Just inline `mountPopper` and `destroyPopper`, you may not even need the instance ref (you can keep it inside the effect scope)

peterhass: Please check if `children` in deps is really needed - we may not need to re-init popper on every re-render.

packages/components/src/base_tooltip/utils.jsx
```diff
@@ -0,0 +1,27 @@
+import {
+  POSITION_TOP,
+  POSITION_BOTTOM,
+  POSITION_LEFT,
+  POSITION_RIGHT,
+  TOOLTIP_OFFSET,
+  TOOLTIP_FLIP_SCREEN_OFFSET,
+} from './constants';
+
+export const getPopperOptions = (ref, position) => ({
```
peterhass, line 10: `ref` -> `arrowRef`

packages/components/src/feature_alert/index.jsx
```diff
@@ -0,0 +1,124 @@
...
+    onOpen,
+    onClose,
+    className,
+    ...cleanProps
+  } = props;
+
+  const [isOpen, setOpen] = useState(defaultOpen);
+  const rootRef = useRef(null);
+
+  const wasActiveOnce = useRef(false);
```
peterhass, line 44: `wasOpenOnce` ?

packages/components/src/feature_alert/index.jsx
```diff
@@ -0,0 +1,124 @@
...
+    </aside>
+  );
+
+  return (
+    <article {...cleanProps} className={rootClassName} ref={rootRef}>
+      <BaseTooltip
+        bubble={bubble} position={position}
+        tooltipClassName={tooltipClassName}
+        arrowClassName={styles.arrow}
+        triggerProps={triggerProps}
```
peterhass, line 82: not needed anymore

packages/components/src/feature_alert/index.jsx
```diff
@@ -0,0 +1,124 @@
...
+    POSITION_BOTTOM,
+    POSITION_LEFT,
+    POSITION_RIGHT,
+  ]),
+  fallbackPosition: PropTypes.oneOf([
+    POSITION_TOP,
+    POSITION_BOTTOM,
+    POSITION_LEFT,
+    POSITION_RIGHT,
+  ]),
```
peterhass, line 110: keep props the same as they previously were - we will add props when we need them

packages/components/src/feature_alert/index.module.scss
peterhass: Yes but visibility has other implications - it's more like 'make this element invisible' - don't know how z-index works in that case

packages/components/src/feature_alert/index.stories.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+let key = true;
+const btnCallback = () => { key = !key; };
+button('Remount tooltip', btnCallback);
+
+export const Default = () => (
+  <div className={styles.container}>
+    <Tooltip
+      content={text('Tooltip text', TOOLTIP_PLACEHOLDER)}
+      position={select('Position', POSITIONING_OPTIONS, POSITION_TOP)}
```
peterhass, line 41: Extract `select('Position', POSITIONING_OPTIONS, POSITION_TOP)` into a variable. I was impressed that it keeps line 50 in sync

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+Tooltip.propTypes = {
+  position: PropTypes.string.isRequired,
+  bubble: PropTypes.oneOfType([
+    PropTypes.object.isRequired,
+    PropTypes.string.isRequired,
+  ]),
+  fallbackPosition: PropTypes.oneOfType([
+    PropTypes.array,
+    PropTypes.string,
+  ]),
```
peterhass, line 41: actually unused - we can get rid of it, not needed at the moment (keep fallback-position functionality, only remove the prop)

packages/components/src/tooltip/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+import styles from './index.module.scss';
+
+const Tooltip = (props) => {
+  const {
+    position,
+    bubble,
+    children,
+    tooltipClassName,
+    arrowClassName,
+    triggerProps,
```
peterhass, line 13: lot of unused stuff here

packages/components/src/tooltip/index.stories.jsx
peterhass: still valid

packages/components/src/tooltip/index.stories.jsx
```diff
@@ -0,0 +1,33 @@
...
+  default: '',
+};
+
+export const Default = () => (
+  <div className={styles.container}>
+    <Tooltip
+      position={select('Tooltip position', POSITIONING_OPTIONS, POSITION_TOP)}
+      bubble={text('Tooltip text', TOOLTIP_PLACEHOLDER)}
+    >
+      {`Tooltip position: ${select('Tooltip position', POSITIONING_OPTIONS, POSITION_TOP)}`}
```
peterhass, line 30: extract `select('Tooltip position', POSITIONING_OPTIONS, POSITION_TOP)` into a variable. I was wondering that it even works this way.

peterhass (changes requested):

packages/components/src/feature_alert/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+    onOpen,
+    onClose,
+    className,
+    ...cleanProps
+  } = props;
+
+  const [isOpen, setOpen] = useState(defaultOpen);
+  const rootRef = useRef(null);
+
+  const wasOpenOnce = useRef(false);
```
peterhass, line 43: did you test this with `defaultOpen=true`?

packages/components/src/feature_alert/index.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const rootClassName = clsx(styles.root, className);
+  const tooltipClassName = clsx(styles.tooltip, { [styles.isActive]: isOpen });
+
+  const triggerProps = getTriggerProps(trigger, handleOpen);
+
+  const bubble = (
+    <aside className={styles.content}>
+      {content}
+      {closeIcon && <span onClick={handleClose} className={styles.cross}><CrossIcon /></span>}
```
peterhass, line 71: if you got time: would `<CrossIcon onClick={handleClose} className={styles.cross} />` work too?

packages/components/src/tooltip/index.jsx
ozerovs: Just can't find this piece of code - super weird

peterhass: I just pulled - should be here https://github.com/goodhood-eu/goodhood/blob/df317c1f55e90108b4f24719630111b4c0127a5d/packages/components/src/tooltip/index.jsx#L11

packages/components/src/base_tooltip/index.jsx
```diff
@@ -0,0 +1,52 @@
...
+  useEffect(() => {
+    const popper = createPopper(
+      content.current,
+      tooltip.current,
+      getPopperOptions(arrow.current, position),
+    );
+    return () => {
+      if (popper) popper.destroy();
+    };
+  }, [content, tooltip, arrow, position]);
```
peterhass, line 30:
```suggestion
  }, [position]);
```

refs don't need to be in deps - they will never change (only the object inside the ref)

peterhass: is `content` needed in deps?

peterhass (approved):

PR #48, opened 2020-11-03
Add browserslist-config-nebenan package

peterhass:


.browserslistrc
```diff
@@ -1,10 +1 @@
...
-
-last 2 major versions
-last 2 years
-> 1%
-not dead
-not op_mini all
-not op_mob > 0
-not ie > 0
-not ie_mob > 0
+extends @goodhood/browserslist-config-nebenan
```
suprMax, line 1: maybe publish as just `browserslist-config-nebenan` and move in a separate repo? this repo holds react ui packages

packages/browserslist-config-nebenan/LICENSE
```diff
@@ -0,0 +1,3 @@
+Copyright (c) Good Hood GmbH.
```
suprMax, line 1: permissive MIT license would work here

.browserslistrc
peterhass: Doesn't really matter imo. Doesn't complicate setup in any way

if you only want to have react stuff in here, then you chose a very generic scope name

peterhass:
Decision: Move to own repo

PR #51, opened 2020-11-06
Change all scss imports to module style imports

peterhass:
There is a bug in `rollup-plugin-postcss` preventing this to work. https://github.com/egoist/rollup-plugin-postcss/issues/333

❌ ~Wait with merge until the bug is fixed~

✅  I published the fix under nebenan-rollup-plugin-postcss and added it to our npm organization. Will revert as soon as they accepted my PR


package.json
```diff
@@ -69,7 +69,7 @@
     "rollup-plugin-delete": "^2.0.0",
     "rollup-plugin-glob-import": "^0.4.5",
     "rollup-plugin-peer-deps-external": "^2.2.3",
-    "rollup-plugin-postcss": "^3.1.8",
```
peterhass, line 72: tbh I cannot find any reference to that version in their repo. Tried to build all our packages and it didn't cause any errors - should be good imo

suprMax: not sure what you mean, why not publish nebenan-rollup-plugin-postcss@1.0.0

peterhass: https://github.com/egoist/rollup-plugin-postcss/blob/master/package.json

no branch in their repo mentions version 3.1.8

I used 3.0.1 to keep their version scheme - would be weird to update later from nebenan-rollup-plugin-postcss 1.0.0 to rollup-plugin-postcss 3.1.x

suprMax: why? it's a different package. it's all good

peterhass: is it worth changing it? would need to unrelease + re-release

suprMax: i think it's better to avoid confision

suprMax (approved):

PR #52, opened 2020-11-10
New logos

suprMax:


packages/components/src/logo/index.jsx
```diff
@@ -3,36 +3,41 @@ import PropTypes from 'prop-types';
 import clsx from 'clsx';
 import styles from './index.module.scss';
 
+import LogoText from '../logo_text';
+
+import { ReactComponent as GHLogo } from './images/logo.svg';
```
peterhass, line 8: `CompanyLogo`

packages/components/src/logo/index.jsx
```diff
@@ -3,36 +3,41 @@ import PropTypes from 'prop-types';
...
+import { ReactComponent as GHChristmasLogo } from './images/logo-christmas.svg';
```
peterhass, line 9: `CompanyChristmasLogo`

packages/components/src/logo/index.jsx
```diff
@@ -3,36 +3,41 @@ import PropTypes from 'prop-types';
...
+  const content = <><LogoImage className={styles.image} />{text}</>;
 
   if (to) return <a {...cleanProps} className={className} href={to}>{content}</a>;
-
   return <span {...cleanProps} className={className}>{content}</span>;
 };
 
 Logo.defaultProps = {
   compact: false,
-  localeName: 'de_de',
```
peterhass, line 30: that's a breaking change imo

suprMax: yes, but i checked and these logos are not in use anywhere, we have to upgrade manually anyway

peterhass: it's still a breaking change

PR #53, opened 2020-11-10
Components: Remove nebenan-components dependency

peterhass:
Current plan for the PR:
- Remove nebenan-components dependency and all of it's peer deps
- Migrate `Markdown` to @goodhood/compoments and use it in CookieBanner

Why remove the `nebenan-components` depdendency?
- It's weird for @goodhood/components to require nebenan-components as a peer dep
- No need to build a mechanism to serve static files yet (emojii assets) (dep was previously missing in components package)


packages/components/src/markdown/index.jsx
```diff
@@ -0,0 +1,84 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import { withRouter } from 'react-router';
+import omit from 'lodash/omit';
+import clsx from 'clsx';
+import marked from 'marked';
+import { invoke } from 'nebenan-helpers/lib/utils';
+import { stripOriginFromUrl } from 'nebenan-helpers/lib/routes';
+import { historyPropTypes } from 'nebenan-react-hocs/lib/history';
+import styles from './index.module.scss';
```
suprMax, line 10: as usual: we will need to be careful migrating to this component. it may cause styling to break

packages/components/src/markdown/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+import { sanitizeText } from './utils';
+
+const EMAIL_LINK_REGEX = /^mailto:/;
+const HTTP_LINK_REGEX = /^https?:/;
+const EXTENSION_REGEX = /\.\w+$/;
+
+const options = { gfm: false };
+
+class Markdown extends PureComponent {
```
suprMax, line 20: now is a good chance to refactor this in a functional way. won't need historyPropTypes/HOC anymore as well

packages/components/src/markdown/index.stories.jsx
```diff
@@ -0,0 +1,25 @@
...
+import longText from './long_text.fixture';
+
+const STORY_PARAMETERS = {
+  knobs: { escapeHTML: false },
+};
+
+export default { title: 'Markdown', component: Markdown, decorators: [withKnobs] };
+
+export const Default = () => (
+  <MemoryRouter>
```
suprMax, line 14: i propose to wrap every story in memory router on storybook level

suprMax (changes requested):

packages/components/src/markdown/index.stories.jsx
peterhass: I propose not to do it. Stories act as usage examples and having the memory router in there shows the dependency to it

packages/components/src/markdown/index.jsx
peterhass: Very good point. Since we don't migrate everything to Markdown now - we should be good but this will definitely break css selectors.

To migrate nebenan-components Markdown usages, we would need to add theme support - should we do it now or later?

suprMax: I think we can get away with root classname here

atomiomi (approved):

packages/components/src/markdown/index.stories.jsx
suprMax: fair point

packages/components/src/markdown/hooks.jsx
```diff
@@ -0,0 +1,32 @@
+import { useCallback } from 'react';
+import { useHistory } from 'react-router';
+import { stripOriginFromUrl } from 'nebenan-helpers/lib/routes';
+import { invoke } from 'nebenan-helpers/lib/utils';
+
+const EMAIL_LINK_REGEX = /^mailto:/;
+const HTTP_LINK_REGEX = /^https?:/;
+const EXTENSION_REGEX = /\.\w+$/;
+
+export const useClickHandler = (onClick) => {
```
suprMax, line 10: imo would benefit from extracting somewhere - we use it in richcontent too

suprMax (approved):

packages/components/src/markdown/hooks.jsx
peterhass: Let's create a favro task for that issue. I am not super confident in extracting that part because it's used over multiple apps and they have slightly different implementations.

https://github.com/goodhood-eu/nebenan-react-hocs/blob/cf03f5bdb4750c7f4390d03b5b744d5ffd30fef0/src/use_smart_click_handler/index.jsx#L8
https://github.com/good-hood-gmbh/nebenan-frontend/blob/cce41446f4f5bc130ffcaa12eae117cde68f40fc/client/ui/rich_content/index.jsx#L34

peterhass: https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-35767

PR #56, opened 2020-11-23
Pkgs upgrades

suprMax:


packages/chargebee/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "1.0.3",
+  "version": "1.1.0",
```
peterhass, line 14: you didn't add anything, therefore patch version bump?

suprMax: i did bump peers, but eh not sure, let's try patch version

peterhass: breaking peer dump -> major
nothing breaking -> patch

why minor?

PR #63, opened 2020-12-04
Map - detect WebGL support

atomiomi:


packages/map/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "2.1.0",
+  "version": "2.2.0",
```
peterhass, line 14: adding a new required prop is a breaking release -> 3.0.0

packages/map/src/map/index.jsx
```diff
@@ -13,6 +13,7 @@ const Map = ({
   className,
   children,
 
+  translations,
```
peterhass, line 16: I would go with `labels` (same as in react datepicker)

packages/map/src/map/index.jsx
```diff
@@ -66,12 +76,14 @@ Map.defaultProps = {
   lockedMobile: true,
   noAttribution: false,
   fitPadding: 20,
+  translations: {},
```
peterhass, line 79: no need for default props if prop is required

packages/map/src/map/index.jsx
```diff
@@ -66,12 +76,14 @@ Map.defaultProps = {
...
 };
 
 Map.propTypes = {
   className: PropTypes.string,
   children: PropTypes.node,
 
+  translations: PropTypes.object.isRequired,
```
peterhass, line 86: we could describe the shape of the object here since we already know it

packages/map/src/map/index.jsx
atomiomi: i used tranlsations in cookiebanner

packages/map/src/map/index.jsx
atomiomi: don't want to pass translation object in every story

packages/map/src/map/index.jsx
atomiomi: would work now. But it can grow. Don't want to write every key in propType check. We don't do it in calendar and cookie banner

packages/map/src/map/index.jsx
suprMax: i second `translations`, clearer

packages/map/src/map/index.jsx
```diff
@@ -53,9 +54,18 @@ const Map = ({
     maxZoom,
   });
 
+  const rootClassName = clsx(styles.root, className, { [styles.isDisabled]: !isWebGLSupported });
+
+  let content;
+  if (isWebGLSupported) {
+    content = <Provider value={context}>{children}</Provider>;
+  } else {
+    content = <h2 className={styles.disabledMessage}>{translations.webgl_disabled}</h2>;
```
suprMax, line 63: h2 doesn't make much sense here, go with something neutral like `div` or `span` (no inherited styles, no semantic value)

packages/map/src/map/index.jsx
suprMax: i would prefer to do something like `const lang = props.translations || defaultTranslations` in render

suprMax (commented):
Given the minuscule userbase that DONT have webgl support i'm not sure we need to even bother with translations tbh.

check comments pls and let's release as non breaking. we could hardcode an english text for now and adjust if PMs/CS complain

packages/map/src/map/hooks.jsx
```diff
@@ -34,12 +34,17 @@ export const useMapEffect = (fn, deps = []) => {
 
 export const useMapInstance = (nodeRef, options) => {
   const [mapInstance, setMap] = useState(false);
+  const [webGLSupport, setWebGLSupport] = useState(true);
```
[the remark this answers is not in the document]
suprMax: also probably setting it to `false` by default and enabling later would be better, example:

```
const useWebGLStatus = () => {
  // Probably setting to `false` by default is better to avoid crashes
  const [isSupported, setSupported] = useState(false);
  useEffect(() => { setSupported(isWebGLSupported(global.document)); }, []);
  return isSupported;
}
```

packages/map/src/map/utils.jsx
```diff
@@ -90,3 +90,11 @@ export const getFitBoundsOptions = ({
 
   return [boundingBox, options];
 };
+
+export const isWebGLSupported = (doc) => {
+  const canvas = doc.createElement('canvas');
+
+  // experimental-webgl for Edge browser
+  const gl = canvas.getContext('webgl') || canvas.getContext('experimental-webgl');
```
suprMax, line 98: did you research what is the best way to check for support? example: https://github.com/Modernizr/Modernizr/blob/acb3f0d990d3424ee7fade52796f9f97920354fc/feature-detects/webgl.js

also checking for support needs to be wrapped in try/catch, otherwise crashes:
https://github.com/Modernizr/Modernizr/blob/acb3f0d990d3424ee7fade52796f9f97920354fc/feature-detects/webgl/extensions.js

suprMax (changes requested):
Won't work in this implementation, see comments

packages/map/src/map/utils.jsx
suprMax: `'WebGLRenderingContext' in window;`

packages/map/src/map/index.jsx
```diff
@@ -66,12 +76,14 @@ Map.defaultProps = {
   lockedMobile: true,
   noAttribution: false,
   fitPadding: 20,
+  webGLError: 'It appears your browser isn\'t supporting WebGL, the HTML standard used to view 3D graphics.',
```
suprMax, line 79: `We couldn't show you the map because current browser doesn't support WebGL.`

PR #68, opened 2020-12-10
Export of variables

ozerovs:


packages/components/src/base_tooltip/index.jsx
```diff
@@ -53,4 +53,5 @@ BaseTooltip.propTypes = {
   arrowClassName: PropTypes.string,
 };
 
+export * from './constants';
```
peterhass, line 56: Please check how constants are named. We export named constants 1:1, so we need easily identifiable and collision free names

peterhass: I think it could make sense to prefix all of them with 'TOOLTIP_'

packages/components/src/base_tooltip/constants.jsx
```diff
@@ -10,4 +10,4 @@ export const TRIGGER_HOVER = 'hover';
 export const TRIGGER_CLICK = 'click';
 export const TRIGGER_DELAYED = 'delayed';
```
peterhass, line 11: if you globally export it -> rename it 
if you don't globally export it -> local naming 

Since you export the whole file -> rename it

peterhass (approved):

PR #73, opened 2020-12-18
Fix tooltip issues again maybe

ozerovs:


packages/components/src/base_tooltip/index.module.scss
```diff
@@ -0,0 +1,14 @@
...
+  // Fix to avoid jumping during tooltip recalculation
+  position: absolute;
+  top: 0;
+  left: 0;
+
+  z-index: -1;
+  visibility: hidden;
+
+  &.isActive {
+    z-index: 100;
```
peterhass, line 11: I'm not sure we need a z-index that high. To be honest, I am still not quite too sure about how z-index works. We use z-index: 9 in modals. I wouldn't want feature alerts overlapping modals. Can you check if something like `3` or `4` would work as well?

packages/components/src/feature_alert/index.jsx
```diff
@@ -76,7 +76,7 @@ const FeatureAlertTooltip = (props) => {
       <BaseTooltip
         bubble={bubble} position={position}
         tooltipClassName={tooltipClassName}
-        arrowClassName={styles.arrow}
+        arrowClassName={styles.arrow} active={isOpen}
```
peterhass, line 79: straighten naming: either use `active` or `open`

packages/components/src/feature_alert/index.module.scss
```diff
@@ -16,17 +16,10 @@ $feature-alert-arrow-size: 16px;
   width: 300px;
   position: absolute;
   cursor: pointer;
-  z-index: -1;
 
   animation-duration: 300ms;
   animation-name: feature-alert-appear;
   animation-timing-function: ease-in;
```
peterhass, line 22: does this thing still work?

packages/components/src/tooltip/index.jsx
```diff
@@ -13,9 +13,14 @@ const Tooltip = (props) => {
     ...cleanProps
   } = props;
 
+  const [isActive, setActive] = useState(false);
+
   return (
     <BaseTooltip
       {...cleanProps}
+      onMouseEnter={() => setActive(true)}
+      onMouseLeave={() => setActive(false)}
```
peterhass, line 22:
```suggestion
      onMouseEnter={setActive.bind(undefined, true)}
      onMouseLeave={setActive.bind(undefined, false)}
```

?

packages/components/src/feature_alert/index.module.scss
ozerovs: It does, why?

packages/components/src/base_tooltip/index.module.scss
ozerovs: Will put 8 for the sake of modals.

peterhass (approved):

PR #75, opened 2021-01-19
Map fixes

atomiomi:


packages/map/src/marker/hooks.js
```diff
@@ -0,0 +1,15 @@
+import { useContext, useEffect } from 'react';
+import MarkerContext from './context';
+
+
+export const useMarkerContext = () => useContext(MarkerContext);
+
+export const useMarkerEffect = (fn, deps = []) => {
+  const marker = useMarkerContext();
+  deps.push(marker);
```
peterhass, line 9: is there any reason why you are mutating the argument? Would prefer `[marker, ...deps]`

packages/map/src/popup/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+Popup.defaultProps = {
+  offsetX: 0,
+  offsetY: 0,
+};
+
+Popup.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.node,
+  defaultOpen: PropTypes.bool,
```
peterhass, line 53: Possible issue: Pretty similar to nebenan-forms Inputs, we are building here an uncontrolled component. I think it's way easier to have controlled components and if you really don't want to handle their state changes, you build a wrapper for it. 

I see myself needing a state controlled popup in the future

packages/map/src/popup/index.stories.jsx
```diff
@@ -0,0 +1,55 @@
...
+    }, 1000);
+
+    return () => clearTimeout(timer);
+  });
+
+  return <strong>{time || 'Loading...'}</strong>;
+};
+
+export const Default = () => {
+  const isActive = boolean('Active', true);
```
peterhass, line 27: why do we need this?

atomiomi: to check that component doesn't throw errors on unmount

packages/map/src/popup/index.jsx
atomiomi: On second thought, this will require handling misclicks as well and we will need to write a wrapper to hold a state because in most cases we don't want to control it.

There is a pretty simple fix if you want to control popup state -- use key prop

suprMax (approved):
LGTM

PR #76, opened 2021-01-20
Add LinkHeader and add new label prop

peterhass:
Added the new `label` prop because I need to build this:

Screenshot 2021-01-20 at 08 59 45

### DIFF
--- a/packages/components/src/link_header/index.jsx
+++ b/packages/components/src/link_header/index.jsx
@@ -9,6 +9,7 @@ const LinkHeader = ({
   children,
   to,
   onClick,
+  label,
   reversed,
   className: passedClassName,
   ...cleanProps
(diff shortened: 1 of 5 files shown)


packages/components/src/link_header/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+  return <Link {...cleanProps} {...{ to, className, onClick }}>{content}</Link>;
+};
+
+LinkHeader.propTypes = {
+  className: PropTypes.string,
+  to: PropTypes.string,
+  onClick: PropTypes.func,
+  reversed: PropTypes.bool,
+  children: PropTypes.node,
+  label: PropTypes.node,
```
peterhass, line 58: `label` is only shown if the component is clickable - suggestions for a better name?

atomiomi (approved):

packages/components/src/link_header/index.jsx
suprMax: `action`

PR #77, opened 2021-01-22
Add theme prop to LinkHeader

peterhass:
I am not really confident into this PR. It's hard to get the design team to generalize stuff.
https://nebenan.slack.com/archives/C01A18L64TV/p1611161946040700?thread_ts=1611153894.038300&cid=C01A18L64TV

Looks like this 'theme' may be used again in the future. I don't want to create a new component that work like this one but only with minor style differences


suprMax (approved):
LGTM

PR #80, opened 2021-01-29
Migrate outside-clicks to eventproxy

peterhass:
eventproxy now handles the react 17 event propagation issue

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -29,7 +29,7 @@
     "date-fns": "^2.16.1",
     "lodash": "^4.17.20",
     "nebenan-form": "^8.9.0",
-    "nebenan-helpers": "^4.5.0",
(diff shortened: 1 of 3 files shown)


atomiomi (approved):

packages/components/package.json
```diff
@@ -29,7 +29,7 @@
     "date-fns": "^2.16.1",
     "lodash": "^4.17.20",
     "nebenan-form": "^8.9.0",
-    "nebenan-helpers": "^4.5.0",
+    "nebenan-helpers": "^4.6.0-beta.0",
```
atomiomi, line 32: why package version is not updated?

peterhass: will do right before merge and publish

suprMax (approved):

PR #81, opened 2021-02-01
Advertisements

suprMax:


packages/components/src/advertisement/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+    window.adn.useCookies(false);
+    window.adn.useLocalStorage(false);
+
+    invoke(onLoad, newUID, window.adn);
+  };
+
+  return (
+    <div className={cx(styles.root, className)}>
+      <Script url={AD_SCRIPT_CDN} onLoad={handleLoad} />
```
peterhass, line 73: let's say I use five `Advertisement` components on one page - does the script get included five times?

packages/components/src/advertisement/utils.jsx
```diff
@@ -0,0 +1,8 @@
+export const getOptions = (object = {}) => (
```
peterhass, line 1: `_.filter(object || {}, (val) => val)` ?

packages/components/src/advertisement/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+      auId: id,
+
+      isolateFrame: true, // improves sandboxing, may cause issues with some ads
+      useCookies: false, // redundant, added just in case
+      protocol: 'https',
+    };
+
+    window.adn.request(requestOptions);
+    invoke(onRequest, uid, requestOptions);
+  }, [uid]); // Intentionally only loads ad once
```
peterhass, line 59: I guess you could add `id`? Doesn't really add complexity - quick win

peterhass, line 59: Let's add all the used props in here. It should work right out of the box. I always expect react components to react to prop changes

packages/components/src/advertisement/index.stories.jsx
```diff
@@ -0,0 +1,14 @@
+import Advertisement from './index';
+
+export default { title: 'Advertisement', component: Advertisement };
+
+export const Default = () => (
+  <Advertisement
+    id="00000000000e25e9"
+    categories={['Brandenburg/Müllrose/15299/Bart-Müllrose']}
```
peterhass, line 8: examples for the other props?

peterhass (approved):

packages/components/src/advertisement/index.stories.jsx
suprMax: not really important as they are not used/solved by adapter and passed into adserver directly. perhaps a link to docu would make sense?

peterhass: yes!

packages/components/src/advertisement/index.jsx
suprMax: react load script thingy manages that exact case

packages/components/src/advertisement/index.jsx
suprMax: won't work here since adserver will load a new ad every time. we need to make it sticky. this just reacts to adscript being loaded basically, the rest is never changing.

packages/components/src/advertisement/utils.jsx
suprMax: only useful for an array of objects, we need to filter object keys 
Screen Shot 2021-02-01 at 4 39 56 PM

peterhass: `_.pickBy(object || {}, (val) => val)`

suprMax: that could fit

peterhass (approved):

PR #82, opened 2021-02-03
Components: Fix cursor for non-clickable LinkHeaders

peterhass:


packages/components/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "2.9.0",
+  "version": "2.9.1",
```
atomiomi, line 14: not updated in lock file

packages/components/package-lock.json
```diff
@@ -13,12 +13,6 @@
         "regenerator-runtime": "^0.13.4"
       }
     },
-    "@goodhood/icons": {
```
atomiomi, line 16: why remove oodhood/icons?

packages/components/package.json
peterhass: we never update the lockfile anyway, it's super annoying we have it anyway. you have to install from root package

packages/components/package-lock.json
peterhass: i assume because we moved it to root package some time ago

suprMax (approved):

PR #85, opened 2021-03-09
Add url and ogSiteName prop to MicroHelmet

peterhass:
https://jira.nebenan.de/browse/CORE-6758

PMs want to be able to change the url on per-site-basis


suprMax (approved):

packages/components/src/micro_helmet/utils.jsx
```diff
@@ -1,18 +1,19 @@
...
-  'url',
-];
+const PROXY_PROPS = {
+  title: 'title',
+  description: 'description',
+  image: 'image',
+  robots: 'robots',
+  canonical: 'canonical',
+  url: 'url',
+  ogSiteName: 'og_site_name',
```
suprMax, line 9: we already had that meta prop and it was og_site_name so you might want to reformat it later when it's passed to the template engine in frontend app

suprMax (approved):

packages/components/src/micro_helmet/utils.jsx
peterhass: In our configs it's called og_site_name and it gets mixed into locals.meta. Didn't want to add a reformatting util for each of our apps. 

But what's your point here? Call it ogSiteName and then change it to og_site_name before passing to the template or calling the prop og_site_name?

suprMax: i'm not sure anymore, i think static meta from phraseapp gets overriden by props from microhelmet so if the prop names match you should get the right one rendering at the right place, not sure why it's working without

PR #88, opened 2021-03-12
remove preflight

suprMax:


packages/components/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "3.1.0",
+  "version": "3.2.0",
```
suprMax, line 14: why regular version bump? npm ignores patch bumps

packages/components/src/advertisement/index.jsx
```diff
@@ -13,25 +13,27 @@ const Advertisement = ({ className, src, children, onRequest, onLoad, ...props }
   const targetClass = `adn-${uid}`;
   const options = getRequestOptions(props);
 
+  // Using the callbacks will flash ad containers briefly when there are no ads to load
+  // This is the desired behavior explicitly asked for by PMs
```
peterhass, line 17: 🤡   ?

packages/components/package.json
peterhass: 🤯

peterhass (approved):

PR #90, opened 2021-04-06
Map: Add Map onBoundsChange prop and Popup onClose, onOpen, maxWidth and closeButton props

peterhass:
https://jira.nebenan.de/browse/CORE-6893


packages/map/src/popup/hooks.jsx
```diff
@@ -0,0 +1,11 @@
+import { useCallback, useEffect, useRef } from 'react';
+
+// TODO: extract to hocs package?
```
peterhass, line 3: https://github.com/goodhood-eu/nebenan-react-hocs/pull/22

packages/map/src/popup/index.jsx
```diff
@@ -45,6 +67,7 @@ const Popup = ({
 Popup.defaultProps = {
   offsetX: 0,
   offsetY: 0,
+  closeButton: true,
```
suprMax, line 70: `closable`?

suprMax (approved):

packages/map/src/popup/index.jsx
peterhass: closeButton only shows the close button itself, the popup is always closable

packages/map/src/map/hooks.jsx
```diff
@@ -64,14 +74,22 @@ export const useMapInstance = (nodeRef, options) => {
       setMap(map);
     };
 
+    const handleBoundsChange = (e) => {
+      const isCausedByUser = Boolean(e.originalEvent);
+      if (!isCausedByUser) return;
+
+      onBoundsChange(map.getBounds().toArray());
```
atomiomi, line 81: Maybe pass bounds instead of array? There can be cases where you will need bounds

atomiomi (approved):

packages/map/src/map/hooks.jsx
peterhass: I thought about it but `bounds` is an object with many methods and I wouldn't want to have our apps rely on implementation details of this package. Let's say we change the underlying map library, then we would need to keep supporting the mapbox `bounds` object api.

suprMax (approved):
SGFMA

PR #91, opened 2021-04-06
Update package lockfiles via root package (no need to install inside package anymore)

peterhass:
Lerna doesn't manage package lockfiles on it's own. This is a workaround so that we don't need to manually run `npm i` for single packages. Will prolong installation time of the root package

https://github.com/lerna/lerna/issues/1462

### DIFF
--- a/package-templates/react/package.json
+++ b/package-templates/react/package.json
@@ -25,7 +25,7 @@
     "start": "../../scripts/start-storybook",
     "start:ssr-preview": "../../scripts/start-storybook --ssr-preview",
     "test": "../../scripts/run-tests",
-    "updatePackageLock": "npm i --package-lock-only"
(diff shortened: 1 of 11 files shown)


package-templates/react/package.json
```diff
@@ -24,7 +24,8 @@
     "build": "rollup --config rollup.config.js",
     "start": "../../scripts/start-storybook",
     "start:ssr-preview": "../../scripts/start-storybook --ssr-preview",
-    "test": "../../scripts/run-tests"
+    "test": "../../scripts/run-tests",
+    "updatePackageLock": "npm i --package-lock-only"
```
peterhass, line 28: `build-package-lock` ?

suprMax: i think we separate words with `:` in npm scripts, so something like `update:lock` would do

suprMax (approved):

PR #92, opened 2021-04-13
Map: Add fancy popup styling

peterhass:
https://github.com/good-hood-gmbh/nebenan-frontend/pull/1313/files#r611775106

### DIFF
--- a/packages/map/package.json
+++ b/packages/map/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "4.2.1",
+  "version": "4.3.0",
   "module": "lib/index.esm.js",
   "main": "lib/index.js",
   "sideEffects": false,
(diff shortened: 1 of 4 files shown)


packages/map/src/popup/index.module.scss
```diff
@@ -0,0 +1,6 @@
+@import "~nebenan-ui-kit";
+
+.root:global(.mapboxgl-popup-content) {
```
atomiomi, line 3: does it work? mapbox usually moves popup up to body node

atomiomi (approved):

packages/map/src/popup/index.module.scss
peterhass: Yes it works because I pass the className into the MapboxPopup constructor

packages/map/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "4.2.0",
+  "version": "4.2.1",
```
suprMax, line 14: not a bugfix go for 4.3.0

suprMax (approved):
SGFMA

packages/map/src/popup/index.module.scss
peterhass: No, it's wrong. Sorry

PR #93, opened 2021-04-20
Migrate to maplibre

atomiomi:


packages/map/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## v5.0.0 Breaking changes
```
peterhass, line 1: why is it a breaking change?

atomiomi: You don't really need to change anything from user perspective but still, we are migrating from one library to another

peterhass: If users don't need to do anything, then it's no breaking change

atomiomi: Hm, I already published the package would be difficult to revert, let's have this version

peterhass: 😭

suprMax: no problem just bump the dependencies in the apps

peterhass: At least add a comment here in CHANGELOG explaining that nothing needs to be changed

PR #103, opened 2021-05-12
Components: Raise CookieBanner to a higher z-index layer

peterhass:
@goodhood/components nebenan-ui-kit peer dep may not need to be updated since the variables are used during compile time. We may only need to update the peer dep if are using new css classes or maybe rely on updated styles(?)

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "5.0.0",
+  "version": "5.1.0",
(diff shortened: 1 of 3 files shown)


packages/components/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "5.0.0",
+  "version": "5.1.0",
```
atomiomi, line 14: maybe patch?

atomiomi (approved):

suprMax (approved):

PR #106, opened 2021-05-25
[Map] Fix sentry error: flat is not a function

atomiomi:


peterhass (approved):

peterhass (approved):
It's fine but I would prefer to fix up the apparently not-working `flat` polyfill

suprMax (approved):

PR #107, opened 2021-06-01
Rename icon groups

peterhass:
https://jira.nebenan.de/browse/CORE-7098


peterhass:
Since we change the api of @goodhood/icons we have to release a new major version.
Since every package using @goodhood/icons lists it as a peer dep, we have to release each of these packages as a new major version.
Every package depending on one of those packages needs a major version bump to support the new version.

@suprMax

PR #112, opened 2021-07-23
Host static preview via github pages

peterhass:
https://favro.com/organization/aee06c3fe6503c804f507dd9/22c630cc02b09f2d6f98b8b3?card=Goo-45453


.github/workflows/preview.yml
```diff
@@ -0,0 +1,23 @@
+name: Build & Deploy preview
+on:
+  push:
+    branches:
+      - main
```
atomiomi, line 5: you can also add paths, to trigger it only when icons were changed
    paths:
      - 'packages/icons/**/*'

PR #114, opened 2021-09-02
Scrollable and Confirm Modal fixes

ozerovs:
- Replaced old Scrollable modal JS-sizes calculation with CSS
- Applied calculation of Scrollable modal sizes based on props
- Added prop for Confirm modal, in order to pass a proper className.


peterhass (approved):
hot 🔥

PR #117, opened 2021-09-08
[CORE-8048, CORE-8049] fix Alert component to support footer button

inomn:
https://jira.nebenan.de/browse/CORE-8048 & https://jira.nebenan.de/browse/CORE-8049

### DIFF
--- a/packages/modals/package.json
+++ b/packages/modals/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "0.5.0-beta.0",
+  "version": "0.5.0",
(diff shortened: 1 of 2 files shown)


packages/modals/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "0.4.0",
```
peterhass, line 14: would publish as 0.4.1 since it's kind of a fix

peterhass (approved):

packages/modals/package.json
inomn: fair enough

PR #120, opened 2021-09-22
Tracking: Listen to event prop changes to prevent component-reuse to stop tracking

peterhass:


eliasbraun (approved):
Looking good to me!

PR #121, opened 2021-09-22
Add side-scroller (& draggable) functional component

eliasbraun:


packages/components/src/constants/swipe.jsx
```diff
@@ -0,0 +1 @@
+export const DISABLE_SCROLL_DISTANCE = 10;
```
ozerovs, line 1: Is it used anywhere besides side_scroller? If not - should be moved to side_scroller folder

packages/components/src/side_scroller/controls.jsx
```diff
@@ -0,0 +1,77 @@
+import clsx from 'clsx';
+import PropTypes from 'prop-types';
+import styles from './controls.module.scss';
+import { getAnimationPosition } from './utils';
+import { SHIFT_PERCENT, SHIFT_TOLERANCE, ANIMATION_DURATION, ANIMATION_FPS } from './constants';
+
+const Controls = ({
+  canScrollLeft,
+  canScrollRight,
+  node,
```
ozerovs, line 10: I think it'll be better to rename it to `scrollable node`

packages/components/src/side_scroller/controls.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const handleScrollLeft = () => {
+    handleShift(-SHIFT_PERCENT);
+  };
+
+  const handleScrollRight = () => {
+    handleShift(SHIFT_PERCENT);
+  };
+
+  const renderControl = (type, handler) => (
```
ozerovs, line 50: Render Control should be a component, instead of render-function. 
Though, it's so simple - I guess we can refactor it like this:

Imo you can extract it in the same file, pass `type` and `handler` as a prop, and then assign it to `leftControl`/`rightControl` variables. 

@peterhass any comments on this?

packages/components/src/side_scroller/controls.module.scss
```diff
@@ -0,0 +1,43 @@
+@import "nebenan-ui-kit";
+
+.control {
+  // rgba(255, 255, 255, 0) is important to fade through white is Safari
```
ozerovs, line 4: This comment doesn't require any action, but good to memorise this kind of things for potential future cross-browser bugs.

packages/components/src/side_scroller/index.stories.jsx
```diff
@@ -0,0 +1,18 @@
+// import { arrayOf } from 'nebenan-helpers/lib/data';
+import SideScroller from './index';
+
+export default { title: 'SideScroller', component: SideScroller };
+
+export const Default = () => (
+  <SideScroller>
+    {/* {arrayOf(50).map((item, index) => (
```
ozerovs, line 8: We need to remove the comment

packages/components/src/side_scroller/index.stories.jsx
```diff
@@ -0,0 +1,18 @@
+// import { arrayOf } from 'nebenan-helpers/lib/data';
```
ozerovs, line 1: We need to remove the comment

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const handleDragStart = (event) => {
+    // Prevents DOM nodes like images from being dragged
+    event.preventDefault();
+    startXRef.current = eventCoordinates(event, 'pageX').pageX;
+    startPositionRef.current = getScrollableNode().scrollLeft;
+    stopScrollAnimation();
+  };
+
+  const handleDragStop = () => { };
```
ozerovs, line 69: It seems like it's unnecessary and can be removed. 
We had this function mentioned in the original component, but it seems like it was not defined, or something.

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+
+  return (
+    <article {...cleanProps} className={className}>
+      <div
+        className={styles.container} ref={containerRef} style={style}
+        onScroll={updateScroll} onLoad={updateSizes}
+      >
+        <Draggable
+          ref={contentRef}
+          onDragStop={handleDragStop}
```
ozerovs, line 119: If you remove handleDragStop - remove this prop as well

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+  const handleDrag = (event) => {
+    const diff = startXRef.current - eventCoordinates(event, 'pageX').pageX;
+    const shift = Math.abs(diff);
+    const newPosition = startPositionRef.current + diff;
+
+    if (shift >= DISABLE_SCROLL_DISTANCE) {
+      preventClickRef.current = true;
+
+      // Prevents vertical scroll on touch devices
+      event.preventDefault();
```
ozerovs, line 80: There's an error in console related to preventing native event happening on usage of this component - we have it in several places, and most probably it's an issue with draggable component, but we need to discover it to be sure.

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+  const containerWidthRef = useRef(null);
+  const contentWidthRef = useRef(null);
+  const animationIdRef = useRef(null);
+  const startXRef = useRef(null);
+  const startPositionRef = useRef(null);
+  const preventClickRef = useRef(null);
+  const stopListeningToResizeRef = useRef(null);
+
+  const className = clsx(styles.root, passedClassName);
+  const cleanProps = props;
```
ozerovs, line 27: Remove this line

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+import clsx from 'clsx';
+import { size, eventCoordinates, stopEvent } from 'nebenan-helpers/lib/dom';
+import eventproxy from 'nebenan-helpers/lib/eventproxy';
+import styles from './index.module.scss';
+
+import Draggable from '../draggable';
+import Controls from './controls';
+import { DISABLE_SCROLL_DISTANCE } from '../constants/swipe';
+
+const SideScroller = ({ className: passedClassName, children, ...props }) => {
```
ozerovs, line 11: replace `...props` with `...cleanProps`

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,140 @@
...
+  const [canScrollLeft, setCanScrollLeft] = useState(false);
+  const [canScrollRight, setCanScrollRight] = useState(false);
+  const [height, setHeight] = useState(0);
+
+  const containerRef = useRef(null);
```
ozerovs, line 16: I think we initialise refs usually without any value - it seems to be working as well. 
I personally prefer to initialise it with value anyway. 

@peterhass any thoughts?

ozerovs (changes requested):

packages/components/src/side_scroller/index.jsx
eliasbraun: `handleDragStop` is passed to `Draggable` component, and in there is used. Maybe we go trough that one again, before I'm deleting it.

packages/components/src/side_scroller/index.jsx
eliasbraun: Relates to what I mentioned in https://github.com/goodhood-eu/goodhood/pull/121#discussion_r714605150, let's go trough this one together!

packages/components/src/side_scroller/index.jsx
eliasbraun: Which one in particular?

packages/components/src/side_scroller/index.jsx
peterhass: ❤️‍🔥 don't care 🔥

packages/components/src/side_scroller/index.jsx
peterhass: 🗑

ozerovs: Let's do it, but eventually it does nothing in there, so it seems to be redundant.

eliasbraun (commented):
Basically finished, small bug remaining: 
When dragging and moving on mobile devices _"Unable to preventDefault inside passive event listener due to target being treated as passive"_ appears in console.

packages/components/src/side_scroller/index.stories.jsx
ozerovs: Still valid :)

ozerovs (changes requested):
Super small changes + required a discussion with @peterhass on the topic of console error. 
Also a TODO for me - check usages of `reset` method original side_scroller component used, and make sure nothing will break, since we got rid of it. 
Besides that - LGTM. 

UPD: Checked usages of side-scroller, couldn't find any component using `reset` method. We'll test it thoroughly during refactoring on platforms, but I guess we are good to go without it.

packages/components/src/draggable/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+    onDrag?.(event);
+  }, { passive: false });
+
+  useEventListener(documentRef, 'mouseup', (event) => {
+    if (!isMouseActive.current) return;
+
+    isMouseActive.current = false;
+    onDragStop?.(event);
+  }, { passive: false });
```
ozerovs, line 53: So the passive is actually needed, and it's not throwing an error that we had? The one related to `event.preventDefault()`?

ozerovs (approved):
LGTM

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+      preventClickRef.current = false;
+    }
+  };
+
+  const handleAnimateScroll = (callback) => {
+    animationIdRef.current = global.requestAnimationFrame(callback);
+  };
+
+  useEffect(() => {
+    stopListeningToResizeRef.current = eventproxy('resize', updateSizes);
```
peterhass, line 95: We have to really watch out for deps here. Using functions inside `useEffect` without specifying them in `deps` can cause weird effects of state being outdated. I don't have a quick solution for refactoring this - would need to spend a bit more time on it.

@ozerovs do you get what I am saying? 


https://overreacted.io/a-complete-guide-to-useeffect/#but-i-cant-put-this-function-inside-an-effect

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+      // Prevents vertical scroll on touch devices
+      event.preventDefault();
+    }
+
+    getScrollableNode().scrollLeft = newPosition;
+  };
+
+  const handleClickCapture = (event) => {
+    if (preventClickRef.current) {
```
peterhass, line 84: flip if - else and make it a guard condition

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+        animationIdRef={animationIdRef}
+        containerWidthRef={containerWidthRef}
+        contentWidthRef={contentWidthRef}
+        onAnimateScroll={handleAnimateScroll}
+      />
+    </article>
+  );
+};
+
+export default SideScroller;
```
peterhass, line 136: propTypes missing

packages/components/src/draggable/index.jsx
peterhass: I think we need it because we pass the event through the callback and it is expected to be active (by how we currently use the component).

packages/components/src/side_scroller/index.jsx
ozerovs: I do get it. 
It seems that solution can be - wrapping all the functions that are being used in useCallback, and pass it as a dependency, no?

peterhass: I am torn a bit here. It currently works because none of those handlers use stuff from state but it will be very easy to introduce bugs.

Suggestion:
- Don't reuse handlers, extract functions. eg `handleScroll` and `handleLoad` (makes it a bit easier to understand and helps with tracking deps)
- All functions used inside `useEffect` need to have `useCallback` (you don't have to change the handlers!)
- Mention these functions you just covered with `useCallback` in effect deps

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+          {children}
+        </Draggable>
+      </div>
+      <Controls
+        canScrollLeft={canScrollLeft}
+        canScrollRight={canScrollRight}
+        scrollableNode={getScrollableNode()}
+        animationIdRef={animationIdRef}
+        containerWidthRef={containerWidthRef}
+        contentWidthRef={contentWidthRef}
```
peterhass, line 129: We cannot pass around so many refs. That somehow defies the whole component based dataflow logic. Can we extract the scrolling into this component? @ozerovs

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,136 @@
...
+  const preventClickRef = useRef(null);
+  const stopListeningToResizeRef = useRef(null);
+
+  const className = clsx(styles.root, passedClassName);
+
+  // Fixes issue with most browsers reducing size of the scrollable element children
+  // as if to compensate for scrollbars, even when they are hidden
+  const style = { height };
+
+  const getScrollableNode = () => containerRef.current;
```
peterhass, line 32: Can we inline this one? Please check if there are any usages of `SideScroller` that imperatively called `getScrollableNode` because otherwise I don't really get why we had it in the first place

packages/components/src/side_scroller/index.stories.jsx
inomn: Delete commented line please :)

packages/components/src/side_scroller/index.module.scss
```diff
@@ -0,0 +1,18 @@
+@import "nebenan-ui-kit";
+
+.root {
+  overflow: hidden;
+  position: relative;
+}
+
+.container {
+  margin-bottom: -30px;
```
inomn, line 9: i personally don't like negative margin usage, isn't there any way to do it without a negative margin?

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,180 @@
...
+  };
+
+  const handleAnimateScroll = (callback) => {
+    animationIdRef.current = global.requestAnimationFrame(callback);
+  };
+
+  const startScrollAnimation = (target) => {
+    let time = 0;
+
+    const animateScroll = () => {
```
inomn, line 96: why do we need this function, if we just calling it right after declaration?

inomn (changes requested):

packages/components/src/constants/swipe.jsx
eliasbraun: `DISABLE_SCROLL_DISTANCE` is used two times in /side_scroller/index. We moved it to /side_scroller to have it all in one place.

packages/components/src/side_scroller/index.stories.jsx
eliasbraun: Done :)

packages/components/src/side_scroller/index.module.scss
eliasbraun: I agree. My approach would be to get rid of margin & padding and set overflow-x to hidden.

inomn (approved):
LGTM

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,180 @@
...
+    containerRef.current.scrollLeft = newPosition;
+  };
+
+  const handleClickCapture = (event) => {
+    if (!preventClickRef.current) return;
+    stopEvent(event);
+    preventClickRef.current = false;
+  };
+
+  const handleAnimateScroll = (callback) => {
```
peterhass, line 89: name wrongly implies that it's a event handler. just inline it - it's not used anywhere else (or move it into `startScrollAnimation` scope)

packages/components/src/side_scroller/index.jsx
```diff
@@ -0,0 +1,180 @@
...
+    animationIdRef.current = global.requestAnimationFrame(callback);
+  };
+
+  const startScrollAnimation = (target) => {
```
peterhass, line 93: cover all non-handler functions inside the component with `useCallback` to prevent accidents when people try to use them in effects

peterhass (approved):
only some knitpicks

PR #122, opened 2021-10-05
On shift side scroller

ozerovs:


packages/components/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "6.2.0",
+  "version": "6.2.1",
```
peterhass, line 14: new feature added -> minor version bump -> 6.3.0

packages/components/src/side_scroller/index.jsx
```diff
@@ -10,7 +10,8 @@ import Controls from './controls';
 import { DISABLE_SCROLL_DISTANCE, SHIFT_PERCENT, SHIFT_TOLERANCE, ANIMATION_DURATION, ANIMATION_FPS } from './constants';
 import { getAnimationPosition } from './utils';
 
-const SideScroller = ({ className: passedClassName, children, ...cleanProps }) => {
+const SideScroller = (props) => {
+  const { className: passedClassName, children, onShift, onDragStart, ...cleanProps } = props;
```
peterhass, line 14: moving the props object explode thingy down here actually makes it harder for IDEs to analyse props. please move it back into line 13

packages/components/src/side_scroller/index.jsx
```diff
@@ -169,6 +173,8 @@ const SideScroller = ({ className: passedClassName, children, ...cleanProps }) =
 SideScroller.propTypes = {
   className: PropTypes.string,
   children: PropTypes.object,
+  onShift: PropTypes.func,
```
peterhass, line 176: I'm a bit unhappy with the prop name because it gives the impression to always be called when the position inside the scroller changes but it's actually only called if user clicks on one of the arrows.

Similar situation here with `onDragStart`. Wouldn't it be better to have something like `onPositionChange` that covers both cases?

peterhass (changes requested):
maybe let's go through it together tomorrow morning

packages/components/src/side_scroller/index.jsx
ozerovs: I thought that in the beginning as well, but we have 2 cases specified in one trigger (because user can interact with the sidescroller two ways - by pressing buttons and sliding, and it is, well, to different things, imo), so it seemed to be a bit more clear. 
On the other hand - yeah, the case is only one. 

And I've just realised that interaction in this case can actually be tracked via clicking on the block itself, without specifiying is it drag or is it shift.

peterhass (approved):

packages/components/package-lock.json
```diff
@@ -0,0 +1,4958 @@
+{
```
inomn, line 1: as i understand we don't need lock file for components, right @peterhass ?
also we are using yarn for this project

packages/components/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "6.2.0",
+  "version": "6.3.0-beta.0",
```
inomn, line 14: do we really want to merge with beta version?

peterhass: no, will update to release version

PR #124, opened 2021-10-20
Modals: Fix missing width limitation in non-scrollable modals

peterhass:
Previously we only set grid rows, therefore limiting the box size.

We forgot about setting a column, therefore there was no limit on box size (it adjusted to the content).


eliasbraun (approved):
LGTM!

PR #125, opened 2021-10-21
Cleaning prop checkout

ozerovs:
Removing a prop that's being passed to a `<span>` element. 
You can check the error via link `/sponsor/create`

### DIFF
--- a/packages/chargebee/src/action/index.jsx
+++ b/packages/chargebee/src/action/index.jsx
@@ -37,7 +37,9 @@ const Action = ({
       if (!disabled) invoke(onCall, instanceRef.current, global.Chargebee);
     };
 
-    node = <span {...rest} onClick={handleClick} />;
+    const { onStep, ...cleanProps } = rest;
+
+    node = <span {...cleanProps} onClick={handleClick} />;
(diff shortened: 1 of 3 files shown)


packages/chargebee/src/action/index.jsx
```diff
@@ -37,7 +37,9 @@ const Action = ({
       if (!disabled) invoke(onCall, instanceRef.current, global.Chargebee);
     };
 
-    node = <span {...rest} onClick={handleClick} />;
+    const { onStep, ...cleanProps } = rest;
```
peterhass, line 40: `onStep` isn't used in this component. where does it come from and why should we omit it here?

peterhass (changes requested):

packages/chargebee/src/action/index.jsx
peterhass: did you already release the version? this PR has been lying around for quite a while @ozerovs

ozerovs: Point is valid, will come back to it once will bump into it again.

PR #131, opened 2021-11-30
Components / Advertisments: Add user session id prop

ozerovs:


packages/components/src/advertisement/utils.jsx
```diff
@@ -11,6 +11,7 @@ export const getRequestOptions = ({
   width,
   height,
   categories,
+  adnuntiusId,
```
peterhass, line 14: `userId` ?

peterhass (approved):

peterhass:
@ozerovs  PR title does not match the change at all

PR #133, opened 2021-12-22
[CORE-8418] Fix bad scroll lock for sequence of modals

peterhass:
[JIRA](https://jira.nebenan.de/browse/CORE-8418)

Issue:
- Render modal a
- Scroll gets locked
- Content gets pushed down since modal is always rendered on the very top of the page
- Render modal b
- Modal a is unmounted, scrolling position is restored
- ???


eliasbraun (approved):
Thought about your comment and how to improve, but didn't find another solution.

PR #134, opened 2022-01-06
[CORE-8559] Add new ecosystem bar

eliasbraun:
See [Jira Link](https://jira.nebenan.de/browse/CORE-8559)

**Notes**
- We found out that we have to keep menu items inside of projects and pass as a prop, because we don’t want to implement anything translation related into goodhood
- Therefore we chose to introduce an new prop `items`

### DIFF
--- a/packages/components/src/ecosystem_bar/index.jsx
+++ b/packages/components/src/ecosystem_bar/index.jsx
@@ -22,10 +22,9 @@ const EcosystemBar = ({ items, onFirstSwipe, onItemClick }) => {
             onClick={() => onItemClick(item.event)}
           >
             <span className={styles.onlyMobile}>
-              {item.mobile[item.key]}
+              {item.mobile}
             </span>
             <span className={styles.notMobile}>
-              {item.other[item.key]}
               {item.others}
             </span>
           </a>
--- a/packages/components/src/ecosystem_bar/index.jsx
+++ b/packages/components/src/ecosystem_bar/index.jsx
@@ -1,0 +1,42 @@
+import clsx from 'clsx';
+import PropTypes from 'prop-types';
+import { useRef } from 'react';
+import { useOnceSwipeTracking, useScrolled } from '@/src/ecosystem_bar/hooks';
+import styles from './index.module.scss';
+
+const EcosystemBar = ({ items, onFirstSwipe }) => {
+  const navRef = useRef(null);
+  const scrolled = useScrolled(navRef);
+
+  useOnceSwipeTracking(navRef, onFirstSwipe);
+
+  return (
+    <ul className={clsx(styles.root, { [styles.scrolled]: scrolled })} ref={navRef}>
+      {items.map((item) => (
+        <li key={item.key} className={styles.item}>
+          <a
+            className={clsx(styles.link, { [styles.isActive]: (item.key === item.project) })}
+            href={item.link}
+            rel="noreferrer noopener"
+            target={item.key === item.project ? '_self' : '_blank'}
+          >
+            <span className={styles.onlyMobile}>
+              {item.mobile[item.key]}
+            </span>
+            <span className={styles.notMobile}>
+              {item.other[item.key]}
+              {item.others}
+            </span>
+          </a>
+        </li>
+      ))}
+    </ul>
+  );
+};
+
+EcosystemBar.propTypes = {
+  items: PropTypes.array.isRequired,
+  onFirstSwipe: PropTypes.func.isRequired,
+};
+
+export default EcosystemBar;
--- a/packages/components/src/ecosystem_bar/index.jsx
+++ b/packages/components/src/ecosystem_bar/index.jsx
@@ -15,10 +15,10 @@ const EcosystemBar = ({ items, onFirstSwipe }) => {
       {items.map((item) => (
         <li key={item.key} className={styles.item}>
           <a
-            className={clsx(styles.link, { [styles.isActive]: (item.key === item.project) })}
+            className={clsx(styles.link, { [styles.isActive]: item.isActive })}
             href={item.link}
             rel="noreferrer noopener"
-            target={item.key === item.project ? '_self' : '_blank'}
+            target={item.isActive ? '_self' : '_blank'}
           >
             <span className={styles.onlyMobile}>
               {item.mobile[item.key]}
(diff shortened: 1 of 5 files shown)


packages/components/src/ecosystem_bar/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+  const scrolled = useScrolled(navRef);
+
+  useOnceSwipeTracking(navRef, onFirstSwipe);
+
+  return (
+    <ul className={clsx(styles.root, { [styles.scrolled]: scrolled })} ref={navRef}>
+      {items.map((item) => (
+        <li key={item.key} className={styles.item}>
+          <a
+            className={clsx(styles.link, { [styles.isActive]: (item.key === item.project) })}
```
inomn, line 18: I think having `project: 'core'` in each menu item is not cool, let's extract `project` to a separate prop?

do we actually need to know what project it is? or we can just add `items[].active: true` to some of the menu items and here change the condition to 
```
[styles.isActive]: item.active
```
?

packages/components/src/ecosystem_bar/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+            href={item.link}
+            rel="noreferrer noopener"
+            target={item.key === item.project ? '_self' : '_blank'}
```
inomn, line 21: same with this condition

packages/components/src/ecosystem_bar/index.module.scss
```diff
@@ -0,0 +1,82 @@
+@import "~nebenan-ui-kit";
+
+// IMPORTANT: global class names dependency
+// 'client/components/layout_header/index.jsx'
```
inomn, line 4: do we have the same component in biz and org?

inomn (changes requested):

packages/components/src/ecosystem_bar/index.jsx
eliasbraun: Since we now are creating the items for each project individually we can also set `isActive` key while doing so. I think I'd prefer that, also the solves the passing of another project prop.

packages/components/src/ecosystem_bar/index.module.scss
eliasbraun: Yes, we do!

inomn (approved):

packages/components/src/ecosystem_bar/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+const EcosystemBar = ({ items, onFirstSwipe, onItemClick }) => {
+  const navRef = useRef(null);
+  const scrolled = useScrolled(navRef);
+
+  useOnceSwipeTracking(navRef, onFirstSwipe);
+
+  return (
+    <ul className={clsx(styles.root, { [styles.scrolled]: scrolled })} ref={navRef}>
+      {items.map((item) => (
```
ozerovs, line 15: We usually put map function into a separate render function, though it won't really improve readability much, but will be consistent though.

packages/components/src/ecosystem_bar/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+        <li key={item.key} className={styles.item}>
+          <a
```
ozerovs, line 17: Maybe update PopupLink component and use it in here?

packages/components/src/ecosystem_bar/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+            className={clsx(styles.link, { [styles.isActive]: item.isActive })}
+            href={item.link}
+            rel="noreferrer noopener"
+            target={item.isActive ? '_self' : '_blank'}
+            onClick={() => onItemClick(item.event)}
```
ozerovs, line 22: By moving to render function, seems like will be a good idea to extract it into a handler function as well inside of a render function. 
This way the function is being recreated on every re-render, and it will slightly impact the performance, as I remember. Would appreciate second opinion from @inomn

packages/components/src/ecosystem_bar/index.module.scss
```diff
@@ -0,0 +1,82 @@
...
+$external-header-panel: ".c-layout_header_panel";
+$external-header-menu: ".c-layout_header-menu-body";
+
+.root {
+  height: 40px;
+  padding: 0 24px;
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+  background: #303030;
```
ozerovs, line 15: I think color should be a constant. 
We don't have exactly #303030, but maybe we can use $color-gray-00 in here instead?

ozerovs (changes requested):

packages/components/src/ecosystem_bar/index.jsx
inomn: Let's use our 'bind-way', to be consistent: https://github.com/good-hood-gmbh/nebenan-frontend/blob/6c232124adf70866f856c04d7b31198b5b86470c/client/components/ecosystem_nav_bar/index.jsx#L30

packages/components/src/ecosystem_bar/index.module.scss
inomn: It's a question for Designers, @eliasbraun if you are motivated, ask Leticia, or just keep #303030 (IMHO)

eliasbraun: Will do!

packages/components/src/ecosystem_bar/index.jsx
eliasbraun: Not sure about that one since nebenan-components is deprecated, I think in this case it's okayish to have it like that. Maybe we can migrate PopupLink to goodhood later on, if find more use cases like these?

packages/components/src/ecosystem_bar/index.module.scss
eliasbraun: Checked with design, you were right Sergei `$color-gray-00` is the color of choice here

eliasbraun: Change of plans: since there are considerations to change the navbar color anyway and Daniel worked on that file and chose #303030, we agreed on sticking to that color for now.

ozerovs: Maybe add a FIXME or a TODO, when the UI kit get updated to update it to a variable?

ozerovs (approved):

PR #136, opened 2022-01-18
[CORE-9080] Change node version to 16 and fix security issues

eliasbraun:
[https://jira.nebenan.de/browse/CORE-9080](https://jira.nebenan.de/browse/CORE-9080)

# Notes

Some notes worth mentioning

- Security: We still have some postcss related issues. Can't do anything about them since the packages we're using are not updating to newest postcss version and the maintainer isn't patching security vulnerabilities on old versions. **Please note that we're not vulnerable to attacks here since we're not piping user input into postcss.**
- npm run build (executed on root level) successful 
- Chargebee package
    - clicked 'checkout button': 
    - console: _Failed to load resource: the server responded with a status of 404 ()_ 
    - Also happening for older node version (fermium)
- Map package: 
    - _{status: 404, url: 'https://api.maptiler.com/maps//style.json?key=', name: 'e', message: ''}_
    - Looks like a missing API key not something node related
    - Also happening for older node version (fermium)
    
Maybe to be 100% sure since this is a very sensitive update, we could pair up and go through it quickly, or you also execute this branch locally.


.preview/build/webpack_config.js
```diff
@@ -17,7 +17,7 @@ const {
   ASSET_REGEX,
   SVG_REGEX,
   SCRIPT_REGEX,
-} = require('./webpack-base');
+} = require('./webpack_base');
```
ozerovs, line 20: I'm wondering, what is the change in here? :D

ozerovs:
Seems like we need versions update in `package.json` as well.

.preview/build/webpack_config.js
peterhass: We updated some linter packages and somehow there is a rule about file naming

peterhass:
> Seems like we need versions update in `package.json` as well.

Why? We're not adding new features, not fixing bugs, etc. Package is and was compatible with node 14 and node 16, all the updated stuff relates to tooling

PR #137, opened 2022-01-20
[CORE-8823] new inputs package

inomn:
[JIRA](https://jira.nebenan.de/browse/CORE-8823)


packages/inputs-tiny/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@goodhood/inputs-tiny",
```
peterhass, line 2: would still opt for `@goodhood/inputs-minimal`, don't like `tiny` as a theme name because ... well ... are the inputs tiny or where does it come from? 

still open for better alternatives

packages/inputs-tiny/.svgo.yml
```diff
@@ -0,0 +1,5 @@
+plugins:
```
peterhass, line 1: Depending on what gets merged first, you might have to update the svgo config to the new format, check this PR: https://github.com/goodhood-eu/goodhood/pull/136

packages/inputs-tiny/src/radio/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+  ...rest
+}) => {
+  const handleChange = (e) => {
+    onChange(value.toString(), e);
+  };
+
+  return (
+    <label className={clsx(styles.label, { [styles.disabled]: disabled })}>
+      <input
+        {...rest}
```
peterhass, line 21: Do we continue with passing `rest` on to the `input`? Makes it hard to refactor later. Are there downsides on not doing it? (data attributes?)

packages/inputs-tiny/src/radio/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+        className={clsx(styles.input, className)}
+        onChange={handleChange}
+        disabled={disabled}
+        type="radio"
+        checked={selectedValue === value.toString()}
```
peterhass, line 26: maybe just pass `checked` as prop instead? https://mui.com/components/radio-buttons/#standalone-radio-buttons

packages/inputs-tiny/src/select/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+import { SIZES_KEYS } from '@/src/constants';
+import Meta from '../textfields_meta';
+import styles from './index.module.scss';
+
+const Select = ({
+  label,
+  error,
+  value = '',
+  options,
+  size = 'medium',
```
peterhass, line 12: use constant here?

packages/inputs-tiny/src/select/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+            value={value}
+            disabled={disabled}
+            {...rest}
+          >
+            {options.map((option) => (
+              <option
+                key={option.value}
+                value={option.value}
+              >
+                {option.key}
```
peterhass, line 62: rather call it `label`? just think if we would want to pass jsx here 

maybe we would want a different api in that case? 

- https://vuetifyjs.com/en/components/selects/#selection vuetify does it like you 
- https://mui.com/components/selects/ mui is very different

packages/inputs-tiny/src/textfield/index.jsx
```diff
@@ -0,0 +1,125 @@
...
+  size = 'medium',
+  className,
+  disableBorder,
+  maxLength,
+  placeholder,
+  disabled,
+  type = 'text',
+  ...rest
+}) => {
+  const [inFocus, setFocus] = useState(false);
```
peterhass, line 26: let's think together if it's ok here to have state (because we wanted to build fully controlled components)

packages/inputs-tiny/src/constants.jsx
```diff
@@ -0,0 +1,7 @@
+export const SIZES = {
+  large: 'large',
+  medium: 'medium',
+  small: 'small',
+};
+
+export const SIZES_KEYS = Object.keys(SIZES);
```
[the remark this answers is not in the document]
inomn: used twice in propTypes, it can be `Object.keys(SIZES)` twice on place in propTypes of components

packages/inputs-minimal/README.md
```diff
@@ -0,0 +1,32 @@
+# `@goodhood/inputs-minimal`
+Nebenan inputs package
+ 
+- [Preview](https://goodhood-eu.github.io/goodhood/package-templates/react/preview/)
```
peterhass, line 4: adjust preview url

packages/inputs-minimal/src/radio/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+      {label && <div className={styles.labelText}>{label}</div>}
+    </label>
+  );
+};
+
+Radio.propTypes = {
+  value: PropTypes.string.isRequired,
+  onChange: PropTypes.func.isRequired,
+  name: PropTypes.string.isRequired,
```
peterhass, line 38: why is name required? you could use this control outside of an form - then there's no point in having a name

packages/inputs-minimal/src/radio/index.stories.jsx
```diff
@@ -0,0 +1,77 @@
...
+          name="example1"
+          checked={value === option.value}
+        />
+      ))}
+    </div>
+  );
+};
+
+export const DefaultSelected = () => {
+  const [value, setValue] = useState('2');
```
peterhass, line 38: how does this work? '2' and 2 have different types

As a user of this library, I would think `Radio` passes the original option value in `onChange`, not a value that might have been casted to string. not sure how other libs are handling this

packages/inputs-minimal/src/select/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+import { ReactComponent as ArrowIcon } from './arrow.svg';
+import Meta from '../textfields_meta';
+import styles from './index.module.scss';
+
+const Select = ({
+  label,
+  error,
+  value = '',
+  options,
+  size = SIZES.medium,
```
peterhass, line 13: Not sure if we have a naming convention for 'enums'. I would have used `Sizes.medium` - have no strong opinion about it though but we should stick to one scheme

peterhass (approved):
looks great, some smaller comments

packages/inputs-minimal/src/radio/index.stories.jsx
inomn: it was an issue in story, radio, and select passing exact value to change handler, without type changes. so in this case there is no radio input with '2' value, so no default selected. story fixed

PR #140, opened 2022-01-26
Components: Fix SideScroller animation executing after unmount

peterhass:
https://sentry.nebenan.de/organizations/sentry/issues/38/events/oldest/?project=4&query=is%3Aunresolved&sort=freq&statsPeriod=14d

How to reproduce:
- [Set animation duration to a very high value](https://github.com/goodhood-eu/goodhood/blob/3477b5e32fc84d504f506e0837c35caff4c09d24/packages/components/src/side_scroller/constants.jsx#L6)
- Visit SideScroller preview
- Click the left arrow
- While the animation is still ongoing, visit another story (needs proper timing but works pretty often)
- -> Crash `Cannot read properties of null (reading 'scrollLeft')`

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "6.6.1",
+  "version": "6.6.2",
   "module": "lib/index.esm.js",
   "main": "lib/index.js",
   "sideEffects": false,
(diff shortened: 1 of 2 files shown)


packages/components/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "6.6.1",
+  "version": "6.6.2",
```
inomn, line 14: current version is 7.0.0, but probably we want to publish this fix for 6.x.x as well

inomn (approved):

PR #144, opened 2022-02-08
[CORE-8823] inputs-minimal v1

inomn:
### Current changes:
- onChange: first param is now `event`, second: pure `value` (needed for Formik)
- new styles for Radio input
- new prop `onBlur` and `onFocus`


packages/inputs-minimal/src/textarea/index.jsx
```diff
@@ -76,14 +80,16 @@ const TextArea = ({
...
 TextArea.propTypes = {
   label: PropTypes.string,
-  placeholder: PropTypes.string,
+  name: PropTypes.string,
   error: PropTypes.string,
   value: PropTypes.string,
   onChange: PropTypes.func.isRequired,
   className: PropTypes.string,
   disableBorder: PropTypes.bool,
   maxLength: PropTypes.number,
```
peterhass, line 89: @inomn maybe we should make this a bit more generic - passing something like `hint`? The input length counter could be a HOC for both TextArea and TextField

packages/inputs-minimal/src/radio_group/index.jsx
```diff
@@ -1,41 +1,59 @@
...
+  </div>
+);
 
 const RadioGroup = ({
   onChange,
-  className,
   items,
   name,
   value,
+  Layout = RadioGroupHorizontalLayout,
```
iamuchejude, line 27: In what case would we need to pass a different layout here aside from `horizontal` or `vertical`?

I propose a different prop, `display` and the value will be `inline` or `block`.  Or we do  `layout` as prop name and acceptable values `horizontal` and `vertical`.

packages/inputs-minimal/src/radio/index.jsx
```diff
@@ -40,6 +51,7 @@ Radio.propTypes = {
   checked: PropTypes.bool,
   className: PropTypes.string,
   disabled: PropTypes.bool,
+  inputProps: PropTypes.any,
```
iamuchejude, line 54:
```suggestion
  inputProps: PropTypes.object,
```

PR #145, opened 2022-02-09
[CORE-8806] Support link clicks on mobile within SideScroller

eliasbraun:
[https://jira.nebenan.de/browse/CORE-8806](https://jira.nebenan.de/browse/CORE-8806)


packages/components/src/side_scroller/index.stories.jsx
```diff
@@ -11,3 +11,14 @@ export const Default = () => (
     />
   </SideScroller>
 );
+
+export const SideScrollerWithLink = () => (
```
peterhass, line 15:
```suggestion
export const WithLink = () => (
```

peterhass (approved):

PR #146, opened 2022-03-09
[CORE-9235] Introduce showNull prop to <ReactionsRow>

eliasbraun:
Introduce possibility to show the number 0 instead of returning null.

### DIFF
--- a/packages/components/src/reactions_row/index.jsx
+++ b/packages/components/src/reactions_row/index.jsx
@@ -11,13 +11,13 @@ const ReactionsRow = ({
   reactions,
   onClick,
   className: passedClassName,
-  showNull = true,
+  withNullableCounter = false,
   ...cleanProps
 }) => {
   const { list: sortedReactionTypes, count } = useReactionData(reactions);
   const reactionTypes = sortedReactionTypes.slice(-limit);
 
-  if (withCounter && count === 0) return showNull ? null : 0;
+  if (withCounter && withNullableCounter && count === 0) return null;
 
   const renderReaction = (reaction) => (
     <ReactionIconBubble
@@ -47,7 +47,7 @@ ReactionsRow.propTypes = {
   onClick: PropTypes.func,
   className: PropTypes.string,
   withCounter: PropTypes.bool,
-  showNull: PropTypes.bool,
+  withNullableCounter: PropTypes.bool,
   limit: PropTypes.number,
 };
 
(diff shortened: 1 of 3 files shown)


packages/components/src/reactions_row/index.jsx
```diff
@@ -11,12 +11,13 @@ const ReactionsRow = ({
...
   onClick,
   className: passedClassName,
+  showNull = true,
   ...cleanProps
 }) => {
   const { list: sortedReactionTypes, count } = useReactionData(reactions);
   const reactionTypes = sortedReactionTypes.slice(-limit);
 
-  if (withCounter && count === 0) return null;
+  if (withCounter && count === 0) return showNull ? null : 0;
```
peterhass, line 20: won't work so easy because you still need to apply all the passed stuff like `onClick`, `passedClassName`

packages/components/src/reactions_row/index.jsx
```diff
@@ -11,12 +11,13 @@ const ReactionsRow = ({
   reactions,
   onClick,
   className: passedClassName,
+  withNullableCounter = false,
```
peterhass, line 14: no need to have a default here, will default to falsey anyway

packages/components/src/reactions_row/index.jsx
```diff
@@ -11,12 +11,13 @@ const ReactionsRow = ({
...
   ...cleanProps
 }) => {
   const { list: sortedReactionTypes, count } = useReactionData(reactions);
   const reactionTypes = sortedReactionTypes.slice(-limit);
 
-  if (withCounter && count === 0) return null;
+  if (withCounter && withNullableCounter && count === 0) return null;
```
peterhass, line 20:
```suggestion
  if (withCounter && !withNullableCounter && count === 0) return null;
```

peterhass (changes requested):

packages/components/src/reactions_row/index.jsx
```diff
@@ -46,6 +47,7 @@ ReactionsRow.propTypes = {
   onClick: PropTypes.func,
   className: PropTypes.string,
   withCounter: PropTypes.bool,
+  withNullableCounter: PropTypes.bool,
```
peterhass, line 50: `withZeroableCounter`

peterhass (approved):

PR #147, opened 2022-03-09
Chargebee tracking

ozerovs:


packages/chargebee/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "1.4.0",
+  "version": "1.4.1",
```
eliasbraun, line 14: Imo it should be v1.5.0 since tracking is a new functionality, not a bugfix!

eliasbraun (approved):
Just a little comment about the version!

packages/chargebee/package.json
peterhass: +1

peterhass (approved):

PR #148, opened 2022-03-22
Goodhood: Add ClickAway component 

peterhass:
Extracted from https://github.com/good-hood-gmbh/business-frontend/pull/596 

Previously we always used `eventproxy('click', cb)` to do outside clicks but it doesn't work inside react portal because of the order of how events are propagated. That's why we have to go the react-only route to check for outside clicks.

Please note that we currently don't gurantee the order in which outslide click handlers are being called. Might need to be changed in the future.

### DIFF
--- a/packages/components/src/click_away/index.jsx
+++ b/packages/components/src/click_away/index.jsx
@@ -1,0 +1,73 @@
+import { createContext, useCallback, useContext, useEffect, useRef } from 'react';
+
+const context = createContext(() => {
+  throw new Error('ClickAwayProvider not set up');
+});
+
+export const ClickAwayProvider = ({ children, ...rest }) => {
+  const isInnerClickMapRef = useRef(new Map());
+  const clickAwayHandlersRef = useRef(new Map());
+
+  const handleClick = () => {
+    const innerClickMap = isInnerClickMapRef.current;
+    const handlersMap = clickAwayHandlersRef.current;
+
+    innerClickMap.forEach((isInnerClick, id) => {
+      if (isInnerClick) return;
+
+      handlersMap.get(id)();
+    });
+
+    innerClickMap.forEach((_, id) => {
+      innerClickMap.set(id, false);
+    });
+  };
+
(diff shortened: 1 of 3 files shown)


eliasbraun (approved):
Tested, looks good on my machine, `ClickAwayProvider not set up`  error also is throwing.

packages/components/src/click_away/index.jsx
```diff
@@ -0,0 +1,73 @@
+import { createContext, useCallback, useContext, useEffect, useRef } from 'react';
+
+const context = createContext(() => {
+  throw new Error('ClickAwayProvider not set up');
```
ozerovs, line 4: Not clear, tbh, why do we need it here?

inomn (approved):

packages/components/src/click_away/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+      <div {...rest} onClick={handleClick}>{children}</div>
+    </context.Provider>
+  );
+};
+
+/**
+ * @param onClickAway function to be called when user clicks outside the 'inner element'
+ * @returns {(function(): void)|*} click handler that needs to be attached to the 'inner element'
+ */
+export const useClickAway = (onClickAway) => {
```
ozerovs, line 58: Feels like we are usually spreading context and hooks into separate files, but in this case seems to be ok.

ozerovs (approved):
LGTM

packages/components/src/click_away/index.jsx
peterhass: Let's say some other component in `@goodhood/components` is using the useClickAway. If you don't specify anything in here, then it will just silently not work. Users of those libraries might not know that they are expected to use `ClickAwayProvider`

packages/components/src/click_away/index.jsx
peterhass: put it here since it's pretty much part of the 'public api' but i have no strong opinion. extra file would be ok as well

PR #149, opened 2022-03-29
[CORE-9284] Change arrows to chevrons and add actual arrows

eliasbraun:
- We need that breaking release because we wrongly gave chevrons the name arrow and now need to add actual arrows
- [Jira Link](https://jira.nebenan.de/browse/CORE-9284)


peterhass (approved):
SGFMA 

sieht gut für mich aus

PR #152, opened 2022-04-20
Added placeholder icon

ozerovs:


packages/icons/src/small/Brackets.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg">
```
peterhass, line 1: In the figma design system the icon is called 'Placeholder' - better to have matching naming
https://www.figma.com/file/PBPtYOEh3AvPce2CsnzcgL/%E2%AC%A1-Design-System?node-id=111%3A1369

PR #155, opened 2022-05-12
inputs-minimal: Add PasswordField component

ozerovs:
+ Minor css fix

No Jira

### DIFF
--- a/packages/inputs-minimal/src/password_field/index.jsx
+++ b/packages/inputs-minimal/src/password_field/index.jsx
@@ -4,9 +4,10 @@ import TextField from '../textfield';
 import omit from 'lodash/omit';
 import Eye1Icon from '@goodhood/icons/lib/medium/eye_1';
 import Eye2Icon from '@goodhood/icons/lib/medium/eye_2';
+import { SIZES_KEYS } from '../constants';
 
-const Password = ({
-  className,
+const PasswordField = ({
+  disabled,
   ...restProps
 }) => {
   const [isVisible, setIsVisible] = useState(false);
@@ -18,15 +19,31 @@ const Password = ({
   const type = isVisible ? 'text' : 'password';
   const Icon = isVisible ? Eye2Icon : Eye1Icon;
 
-  const cleanProps = omit(restProps, 'attachmentLeft', 'attachmentRight');
+  const cleanProps = omit(restProps, 'attachmentLeft');
 
   return (
-    <TextField {...cleanProps} __type={type} attachmentRight={<Icon onClick={handleClick} />} />
+    <TextField
+      {...cleanProps}
+      disabled={disabled}
+      __type={type}
+      attachmentRight={<Icon onClick={disabled ? undefined : handleClick} />}
+    />
   );
 };
 
-Password.propTypes = {
+PasswordField.propTypes = {
+  label: PropTypes.string,
+  error: PropTypes.string,
+  value: PropTypes.string,
+  name: PropTypes.string,
+  onChange: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(SIZES_KEYS),
   className: PropTypes.string,
+  disableBorder: PropTypes.bool,
+  hint: PropTypes.string,
+  disabled: PropTypes.bool,
+  onFocus: PropTypes.func,
+  onBlur: PropTypes.func,
 };
 
-export default Password;
+export default PasswordField;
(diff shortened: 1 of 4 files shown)


eliasbraun (changes requested):

packages/inputs-minimal/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "1.0.0",
+  "version": "1.0.1",
```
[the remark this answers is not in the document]
ozerovs: True, makes sense.

eliasbraun (approved):

packages/inputs-minimal/src/textfield/index.jsx
```diff
@@ -20,6 +20,7 @@ const TextField = ({
   disabled,
   onFocus,
   onBlur,
+  type,
```
peterhass, line 23: Maybe add a `Password` (I'm assuming that's what you're using it for) instead of having type here? The type attribute can change a lot about how textfields behave or look like. Let's say you are passing `number` here - would the attachments still work? You would pretty much need to check for every type there is. That's why I think it's better to have separate components instead of this `type` prop -> more control

peterhass: Maybe call the prop `__type` to mark it for internal use and then simply use it in the `Password` component?

packages/inputs-minimal/src/password_field/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+    'icon-eye': !isVisible,
+    'icon-eye_crossed': isVisible,
+  });
+
+  const cleanProps = omit(restProps, 'attachmentLeft', 'attachmentRight');
+
+  return (
+    <span className={clsx(styles.root, className)}>
+      <TextField className={styles.input} {...cleanProps} __type={type} />
+      <i className={iconClassName} onClick={handleClick} />
```
peterhass, line 30: can't you do the same with `attachmentRight`?

packages/inputs-minimal/src/password_field/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+  const handleClick = (event) => {
+    event.preventDefault();
+    setIsVisible((prevState) => !prevState);
+  };
+
+  const type = isVisible ? 'text' : 'password';
+  const iconClassName = clsx(styles.icon, {
+    'icon-eye': !isVisible,
+    'icon-eye_crossed': isVisible,
```
peterhass, line 22: better to use @goodhood/icons - ui-kit iconsfont is deprecated

packages/inputs-minimal/src/password_field/index.stories.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <div className={styles.container}>
+      <Password
+        value={value}
+        label="My Awesome Input"
+        onChange={handleChange}
+        disabled={boolean('Disabled', false)}
+        error={text('Error message', '')}
+        hint={text('Hint', '')}
```
peterhass, line 24: same with hint

packages/inputs-minimal/src/password_field/index.stories.jsx
```diff
@@ -0,0 +1,78 @@
...
+        disableBorder={boolean('disableBorder', false)}
+        size={select('Size', Sizes, Sizes.medium)}
```
peterhass, line 26: icon not vertically centered for every supported size

peterhass (changes requested):

packages/inputs-minimal/src/textfield/index.jsx
```diff
@@ -20,6 +20,7 @@ const TextField = ({
   disabled,
   onFocus,
   onBlur,
+  __type,
```
peterhass, line 23:
```suggestion
  __type = 'text',
```

remove `defaultProps`

packages/inputs-minimal/src/password_field/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+import { SIZES_KEYS } from '../constants';
+
+const PasswordField = ({
+  disabled,
+  ...restProps
+}) => {
+  const [isVisible, setIsVisible] = useState(false);
+
+  const handleClick = (event) => {
+    event.preventDefault();
```
peterhass, line 16: why do we need preventDefault here?

peterhass (approved):
check another time in all browsers - then we're good to go 🚀

PR #158, opened 2022-05-24
[CORE-9919] Migrate progress bar from nebenan components

eliasbraun:
I want to migrate the old progress bar from nebenan components. 
Mainly because I want to make it more customizable, meaning I am able to style the background color of the bar and the progress indicator bar in the foreground independent from each other in an easy way. 
Functionality stays (nearly) the same, but instead of passing a "type" prop, I want to pass 2 className props, which can style the 2 progress bars to make it more flexibel.

### DIFF
--- a/packages/components/src/progress_bar/index.jsx
+++ b/packages/components/src/progress_bar/index.jsx
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import clsx from 'clsx';
-import { getPercentage } from './utils';
+import { getPercentageString } from './utils';
 import styles from './index.module.scss';
 
 const ProgressBar = ({
@@ -11,7 +11,7 @@ const ProgressBar = ({
   progressBarProgressIndicatorStyles,
   ...cleanProps
 }) => {
-  const percent = getPercentage(state);
+  const percent = getPercentageString(state);
 
   const className = clsx(styles.root, progressBarStyles, {
     [styles.small]: size === 'small',
@@ -23,7 +23,7 @@ const ProgressBar = ({
         className={clsx(styles.progressState, progressBarProgressIndicatorStyles)}
         style={{ width: percent }}
       >
-        <em>{percent}</em>
+        <em className={styles.percent}>{percent}</em>
       </span>
       {children}
     </div>
--- a/packages/components/src/progress_bar/index.jsx
+++ b/packages/components/src/progress_bar/index.jsx
@@ -10,7 +10,6 @@ const ProgressBar = ({
   state,
   size,
   theme = PROGRESS_BAR_THEME_MODERN,
-  ...cleanProps
 }) => {
   const percent = getPercentageString(state);
 
@@ -19,7 +18,7 @@ const ProgressBar = ({
   });
 
   return (
-    <div {...cleanProps} className={className}>
+    <div className={className}>
       <span
         className={clsx(styles.progressState, styles[`theme-${theme}`])}
         style={{ width: percent }}
--- a/packages/components/src/progress_bar/index.jsx
+++ b/packages/components/src/progress_bar/index.jsx
@@ -1,0 +1,44 @@
+import PropTypes from 'prop-types';
+import clsx from 'clsx';
+import { getPercentage } from './utils';
+import styles from './index.module.scss';
+
+const ProgressBar = ({
+  state,
+  size,
+  children,
+  progressBarStyles,
+  progressBarProgressIndicatorStyles,
+  ...cleanProps
+}) => {
+  const percent = getPercentage(state);
+
+  const className = clsx(styles.root, progressBarStyles, {
+    [styles.small]: size === 'small',
+  });
+
+  return (
+    <div {...cleanProps} className={className}>
+      <span
+        className={clsx(styles.progressState, progressBarProgressIndicatorStyles)}
+        style={{ width: percent }}
+      >
+        <em>{percent}</em>
+      </span>
+      {children}
+    </div>
+  );
+};
+
+ProgressBar.propTypes = {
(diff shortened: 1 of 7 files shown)


packages/components/src/progress_bar/index.jsx
```diff
@@ -0,0 +1,44 @@
+import PropTypes from 'prop-types';
+import clsx from 'clsx';
+import { getPercentage } from './utils';
+import styles from './index.module.scss';
+
+const ProgressBar = ({
+  state,
+  size,
+  children,
+  progressBarStyles,
```
ozerovs, line 10: `className` maybe?

packages/components/src/progress_bar/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+  progressBarProgressIndicatorStyles,
```
ozerovs, line 11: `progressIndicatorClassName` maybe?

packages/components/src/progress_bar/index.module.scss
```diff
@@ -0,0 +1,40 @@
...
+.progressState {
+  height: 100%;
+  position: absolute;
+  background: $color-green-02;
+  top: 0;
+  left: 0;
+
+  transition: width $transition;
+
+  em {
```
ozerovs, line 27: Would prefer adding a class to the `em` element - won't be a need to nest it then

packages/components/src/progress_bar/utils.jsx
```diff
@@ -0,0 +1,9 @@
+export const getPercentage = (input) => {
+  let percent;
+  if (typeof input === 'number' && input < 1) percent = String(input * 100);
```
ozerovs, line 3: Maybe to simplify it a bit?

```suggestion
  if (typeof input === 'number' && input < 1) return `${input}%`;
  if (input.includes('%')) return String(input);
  return `${input}%`;
```

packages/components/src/progress_bar/utils.jsx
```diff
@@ -0,0 +1,9 @@
+export const getPercentage = (input) => {
```
ozerovs, line 1: `getPercentageString` maybe? 

Also, tests would be good for this one, to illustrate the possible cases that can occur for this util.

ozerovs (changes requested):

ozerovs:
I'd also invite @peterhass to take a look, since he doesn't like passing to much of classes from above. (but I agree - sometimes I personally do it with no valid reason)

packages/components/src/progress_bar/index.jsx
eliasbraun: Because I think `className` and `progressIndicatorClassName` wouldn't really reflect it since `className` is very generic. Nevertheless I agree, my names are pretty bad

peterhass: Why do we need those className props anyway? Any chance you could provide links / screenshots?

eliasbraun: To provide a way to style the 2 progress lines. In this case lightblue and darkblue with border-radius.  



If nothing is passed, the default looks are applied, but if I wouldn't pass them, I style these child elements in a clean way as far as I'm concerned. 
Hope that clears it up.

peterhass: https://github.com/goodhood-eu/goodhood/pull/158#issuecomment-1135985331 @ozerovs  is very correct about me not liking to pass too many classNames :D 

maybe introduce a `theme` prop? 

you could predefine two themes:
- legacy (legacy.module.scss)
- modern (modern.module.scss)

Then it should be a lot easier to refactor in the future. With passing in class names you never know what will break if you change one single css prop

packages/components/src/progress_bar/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+        <em className={styles.percent}>{percent}</em>
+      </span>
+      {children}
+    </div>
+  );
+};
+
+ProgressBar.propTypes = {
+  progressBarStyles: PropTypes.string,
+  progressBarProgressIndicatorStyles: PropTypes.string,
```
peterhass, line 36: 👀

packages/components/src/progress_bar/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  state: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
+  size: PropTypes.string,
+  children: PropTypes.node,
+  theme: PropTypes.object,
```
peterhass, line 43: it's actually a string

packages/components/src/progress_bar/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+import clsx from 'clsx';
+import { getPercentageString } from './utils';
+import styles from './index.module.scss';
+
+export const THEME_MODERN = 'modern';
+
+const ProgressBar = ({
+  state,
+  size,
+  children,
```
peterhass, line 11: why do we even need children here?

packages/components/src/progress_bar/index.jsx
```diff
@@ -0,0 +1,46 @@
+import PropTypes from 'prop-types';
+import clsx from 'clsx';
+import { getPercentageString } from './utils';
+import styles from './index.module.scss';
+
+export const THEME_MODERN = 'modern';
```
peterhass, line 6: 1. missing constant for THEME_LEGACY
2. watch out: named constants will be exported 1:1 to the final esm bundle. Better name it something like `PROGRESS_BAR_THEME_MODERN` - check other named exports in components package for more examples

peterhass (changes requested):

PR #161, opened 2022-06-28
[CORE-10123] Add ProgressLine component and update RadioGroup component

iamuchejude:
Relates to [CORE-10123](https://jira.nebenan.de/browse/CORE-10123) from https://github.com/good-hood-gmbh/nebenan-frontend/pull/1674


packages/components/preview/index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>packages/components</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1"><script defer src="/goodhood/packages/components/preview/main.js"></script><link href="/goodhood/packages/components/preview/main.css" rel="stylesheet"></head>
+  <body>
+  </body>
+</html>
```
zeitwidrig, line 9: I don't get why suddenly it's required to include a preview folder in the components package.
And as far as I understand it's not even used. Instead the root preview folder is used.

peterhass: I think @iamuchejude accidentally commited the preview folder generated by `npm run tryBuild:preview`

packages/inputs-minimal/src/radio_group/index.jsx
```diff
@@ -24,33 +25,42 @@ const RadioGroup = ({
...
+            label={item.label}
+            value={item.value}
+            onChange={handleChange}
+            disabled={item.disabled}
+            checked={value === item.value}
+          />
+        ))}
+      </Layout>
+
+      <Meta
```
peterhass, line 51: I would suggest to put `Meta` into `Layout`. Based on the naming Layout should decide on the whole layout of the component, not only the radios.

Also I think it might be beneficial to have them all under one root element to help with positioning (`TextField` also has a root div)

packages/components/src/progress_line/index.stories.jsx
```diff
@@ -0,0 +1,7 @@
+import ProgressLine, { PROGRESS_LINE_WEIGHT_THIN } from './index';
+
+export default { title: 'ProgressLine', component: ProgressLine };
+
+export const Default = () => <ProgressLine steps={6} current={2} />;
```
peterhass, line 5: rather use knobs to make the story more interactive ✨

packages/components/src/progress_line/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+const ProgressLine = (props) => {
+  const { steps, current, weight, ...cleanProps } = props;
+  const className = clsx(styles.root, props.className, {
+    [styles.thin]: weight === PROGRESS_LINE_WEIGHT_THIN,
+  });
+
+  const chunks = Array.from({ length: steps }, (_, index) => (
+    <li
+      key={index} className={clsx(styles.step, {
```
ozerovs, line 16: Even though I can't think of a better option for key, maybe it's worth prefixing it with some string? 

Not 100% sure how it'll behave if we'll have other components with same index. 

Plus, would also move the className into a variable outside of JSX

packages/components/src/progress_line/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+        [styles.active]: index + 1 <= current,
+      })}
+    />
+  ));
+
+  return (
+    <ul {...cleanProps} className={className}>{chunks}</ul>
```
ozerovs, line 23: Even though it seems like it works, we are just usually going with `.map()` and render function, maybe worth keeping this way?

packages/components/src/progress_line/index.jsx
```diff
@@ -0,0 +1,43 @@
+import PropTypes from 'prop-types';
+import clsx from 'clsx';
+import styles from './index.module.scss';
+
+export const PROGRESS_LINE_WEIGHT_THIN = 'thin';
+export const PROGRESS_LINE_WEIGHT_NORMAL = 'normal';
+
+const ProgressLine = (props) => {
+  const { steps, current, weight, ...cleanProps } = props;
```
ozerovs, line 9: Maybe worth inlining props?

packages/inputs-minimal/src/radio_group/index.jsx
```diff
@@ -24,33 +25,42 @@ const RadioGroup = ({
...
-          onChange={handleChange}
-          name={name}
-          label={item.label}
-          value={item.value}
-          disabled={item.disabled}
-          checked={value === item.value}
-        />
-      ))}
-    </Layout>
+    <div>
```
ozerovs, line 36: I think if comment bellow will be implemented - there won't be a need for div. 
If not - isn't it worth having a Fragment instead? `<>...</>`

packages/components/src/progress_line/index.jsx
peterhass: maybe you could use `arrayOf` here? https://github.com/goodhood-eu/nebenan-helpers/blob/master/src/data/index.js#L154

packages/inputs-minimal/src/radio_group/index.jsx
iamuchejude: https://github.com/goodhood-eu/goodhood/pull/161#discussion_r911696546

packages/inputs-minimal/src/radio_group/index.jsx
peterhass: @iamuchejude no idea why you resolved the conversation because you neither responded nor did the suggested change

iamuchejude: I wrapped the components in a root element as recommended. Placing  the `<Meta>` in `<Layout>` because the layout will align meta with the radios. For example `<HorizontalLayout>` will align all including meta horizontally.

peterhass: The thing is, `RadioGroup` had no `className` prop because the `Layout` prop already takes care of all the placing. There even is already a usage of it.

https://github.com/good-hood-gmbh/business-frontend/blob/b3278f945479fe03e140353bd52beca56850723e/client/containers/poi_profile/components/poi_call_to_action/components/action_form/action_type.jsx#L7

With 'taking it into the Layout' I meant something like a render prop.

`<Layout meta={<Meta />}>radio1 radio2 radio3</Layout>`

Now it's a bit weird because you can pass in a Layout component but still can't control the root element or where the errors are placed. It's actually not the component layout anymore but only `ButtonsLayout`.

iamuchejude: Yeah, passing as a render prop makes sense. I'm just gonna create a new PR with the updates.

PR #164, opened 2022-07-14
inputs-minimal: Prevent label selection in inputs

peterhass:
Text selection shouldn't work on inputs labels.

Please check in preview if it really works. Didn't update versions because it's a very small change and not really important. Just noticed it on my own.


zeitwidrig (approved):
I don't find it very crucial but it does work as intended.

peterhass:
@eliasbraun 

> get rid of the yarn lock file?

I think somebody might have forgotten to commit their lockfile, that's why I included it here.

> Bump the version to 1.3.1?

Didn't update versions because it's a very small change and not really important.

PR #167, opened 2022-08-11
[CORE-10707] Refactor TextFieldMeta component to support FieldGroup

iamuchejude:
Relates to https://github.com/good-hood-gmbh/nebenan-frontend/pull/1718

### DIFF
--- a/packages/inputs-minimal/src/textfields_meta/index.jsx
+++ b/packages/inputs-minimal/src/textfields_meta/index.jsx
@@ -7,10 +7,10 @@ const Meta = ({
   hint,
   className,
 }) => {
-  const hasHintOrError = Boolean(hint || error);
-  const useInputGroup = hasHintOrError && typeof error !== 'string';
+  const hasError = typeof error === 'string';
+  const hasHint = Boolean(hint);
 
-  if (!hasHintOrError || useInputGroup) return null;
+  if (!hasError && !hasHint) return null;
 
   return (
     <div className={clsx(styles.meta, className)}>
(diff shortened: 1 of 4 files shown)


packages/inputs-minimal/src/textfields_meta/index.jsx
```diff
@@ -6,19 +6,22 @@ const Meta = ({
   error,
   hint,
   className,
-}) => (
-  <div className={clsx(styles.meta, className)}>
-    {error && (
+}) => {
+  const hasHintOrError = Boolean(hint || error);
+  const useInputGroup = hasHintOrError && typeof error !== 'string';
```
peterhass, line 11: `input-minmal` has no idea what an 'input group' is. It doesn't exist in the package. Maybe just go with ...

```js
const hasError = typeof error === 'string';
const hasHint = Boolean(hint);

if (!hasError && !hasHint) return null;
```

packages/inputs-minimal/src/textfields_meta/index.module.scss
```diff
@@ -2,6 +2,7 @@
 
 .meta {
   display: flex;
+  font-weight: $font-weight-l;
```
peterhass, line 5: is this in the design system?

peterhass (changes requested):

packages/inputs-minimal/src/textfields_meta/index.jsx
iamuchejude: Yeah, that makes sense.

packages/inputs-minimal/src/textfields_meta/index.module.scss
iamuchejude: Yeah, it is.

peterhass (approved):

peterhass:
Will merge this PR because you already published a stable version anyway (and we're publishing-blocked otherwise)

PR #168, opened 2022-08-11
CORE-10135 Add support for e2e test ids

peterhass:
https://jira.nebenan.de/browse/CORE-10135


packages/inputs-minimal/src/utils.jsx
```diff
@@ -0,0 +1,8 @@
+export const pickDataAttributes = (props) => (
+  Object.keys(props)
```
peterhass, line 2: didn't want to add lodash as dep only for this little thing

PR #172, opened 2022-10-08
[CORE-11289] Fix "undefined source cashes"

rejas:
fixed in 2.1.8-pre.2 of maplibre-gl-js
see https://github.com/maplibre/maplibre-gl-js/pull/1099

### DIFF
--- a/packages/map/package.json
+++ b/packages/map/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "7.0.1",
+  "version": "7.0.2",
   "module": "lib/index.esm.js",
   "main": "lib/index.js",
   "sideEffects": false,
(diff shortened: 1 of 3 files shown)


packages/map/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "7.0.1",
+  "version": "7.0.2-beta.0",
```
zeitwidrig, line 14: Personal opinion: I would never commit beta versions, only the final version. Even when publishing intermediate beta versions to npm.
And with [yalc](https://wiki.nebenan.de/display/AWSM/How+to+work+with+1st+party+packages) this might not even be necessary.

rejas: So how could I get this PR into the Core through QA without changing anythign at all after QA and without releasing an untested version of this package? Even if I dont change the version here and use the branch-name in the dependencies of the nebenan-core, I would have to commit/change something in Core after QA tested the core-branch (since there would be the master branch name used and not the final version name)

rejas:
> I'm not sure if we'd need that yarn.lock as part of the PR?

yarn.lock should always get updated when you update a dependency. As for yarn being used here: Once upon a time yarn was better at handling mono repos (so I was told). Not sure if that is still the case but I dont want to change a running system :-)

PR #173, opened 2022-10-20
[CORE-11160] Migrate @goodhood/modals to TypeScript

iamuchejude:
[JIRA Link](https://jira.nebenan.de/browse/CORE-11160)

Depends on https://github.com/goodhood-eu/styleguide/pull/31


packages/modals/package.json
```diff
@@ -24,6 +25,7 @@
   "scripts": {
     "reinstall": "rm -rf node_modules && yarn install",
     "prepublishOnly": "(cd ../../ && npm run check) && npm run build",
+    "watch": "rollup --config rollup.config.js --watch",
```
peterhass, line 28: why do we need watch scripts? for local dev you would rather use the preview

packages/modals/package.json
```diff
@@ -36,16 +38,17 @@
...
+    "react-dom": "^16.14.0 || ^17.0.1"
   },
   "devDependencies": {
-    "prop-types": "^15.7.2",
+    "@babel/preset-typescript": "^7.18.6",
     "@goodhood/components": "^7.0.0",
     "nebenan-keymanager": "^1.0.1",
-    "nebenan-ui-kit": "^5.3.0"
+    "nebenan-ui-kit": "^5.3.0",
+    "prop-types": "^15.7.2"
```
peterhass, line 52: maybe we can get rid of prop types now?

peterhass (approved):
add linting and we're good to go :)

packages/modals/src/scrollable_modal/index.stories.tsx
```diff
@@ -27,7 +27,7 @@ export const Default = () => {
           staticPosition={staticPosition}
         >
           <ul>
-            {[...(new Uint8Array(numberOfItems))].map((value, key) => (
+            {[...(new Uint8Array(numberOfItems))].map((_, key) => (
```
rejas, line 30: do we define that _ as argsIgnorePattern or is it from airbnb?

iamuchejude: It's actually from airbnb. I agree with you, it's a non-descriptive.  I should just rename from `_` to `_value`, that's prolly clearer.

package.json
```diff
@@ -92,15 +94,14 @@
     "webpack-merge": "^5.7.3"
   },
   "scripts": {
-    "postinstall": "lerna bootstrap && link-parent-bin && link-parent-bin -c package-templates && yarn run tryBuild",
```
peterhass, line 95: we really need to have this `postinstall` script here, otherwise many of internal dependencies break very easily. you can track down when it was added, maybe there's a better reason why but still - very sure we should keep it

packages/modals/src/modal/index.stories.tsx
```diff
@@ -34,7 +34,7 @@ export const Default = () => {
 };
 
 export const LongAndScrollable = () => {
-  const [isActive, setActive] = useState();
+  const [isActive, setActive] = useState < boolean >(false);
```
peterhass, line 37:
```suggestion
  const [isActive, setActive] = useState<boolean>(false);
```

packages/modals/src/modal/index.tsx
```diff
@@ -1,37 +1,61 @@
...
+  useTrack,
   useUnmount,
 } from './hooks';
 import { useModalProvider } from '../provider/hooks';
 import Portal from '../portal';
 
 import styles from './index.module.scss';
 
-const Modal = forwardRef(({
+export type ModalProps = PropsWithChildren<HTMLAttributes<HTMLDivElement> & {
```
peterhass, line 27: `HTMLDivElement` locks us in to `div` as root element. Changing the root node type would cause a breaking change in types

packages/modals/src/modal/utils.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type TrackFunc = (event: TrackEvent) => void;
+
+let trackFunc: TrackFunc;
+
+export const configure = (options: {
+  track?: TrackFunc
+}) => {
+  const { track } = options;
+  if (track === undefined) return;
```
peterhass, line 14: 1. breaking change to how the function works
2. no way to reset `trackFunc` now
3. why?

packages/modals/src/provider/index.tsx
```diff
@@ -1,19 +1,28 @@
...
 import { useProviderValue } from './hooks';
 
+type ModalProviderProps = {
+  className?: string,
+} & Partial<{
+  [key: string]: unknown;
+}>;
 
-const ModalProvider = forwardRef(({
+const ModalProvider: React.FC = forwardRef<unknown, ModalProviderProps>(({
```
peterhass, line 18: wonder why we have to specify `React.FC` here when we never do it anywhere else

iamuchejude: Definitely shouldn't be `React.FC`. My bad.

PR #176, opened 2022-11-08
Prevent event default only if event is cancelable

iamuchejude:
Screenshot 2022-11-08 at 08 48 50


peterhass:
Does that change anything on UI side? any behavior?

iamuchejude:
@peterhass No visible UI change.

rejas (approved):

PR #181, opened 2022-12-07
Keep default mapbox link styles

eliasbraun:
To prevent global link styles from overwriting mapbox links, we need to set them in there.

Issue: https://miro.com/app/board/uXjVO6lyWnY=/?moveToWidget=3458764540076530305&cot=14

### DIFF
--- a/packages/map/src/map/index.module.scss
+++ b/packages/map/src/map/index.module.scss
@@ -16,7 +16,7 @@
     // overrides mapbox z-index
     z-index: 1;
 
-    // keep default mapbox link styles under any circumstances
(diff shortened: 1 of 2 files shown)


packages/map/src/map/index.module.scss
```diff
@@ -15,6 +15,18 @@
   :global(.mapboxgl-ctrl-bottom-right) {
     // overrides mapbox z-index
     z-index: 1;
+
+    // To prevent global link styles from overwriting, set default styles explicitly
```
peterhass, line 19: 😰

rejas (approved):
Looks ok

PR #183, opened 2022-12-15
[CORE-10528] Fix error color for input fields

rejas:
[JIRA Link](https://jira.nebenan.de/browse/CORE-10528)


eliasbraun (changes requested):
Assuming that these are all the places we want to have the color update, then I guess that should be fine! 
One thing missing tho is the package version bump!

eliasbraun (approved):

rejas:
Bumped version. Any missed style changes will be done in https://jira.nebenan.de/browse/CORE-11662

PR #184, opened 2023-01-04
[CORE-11662] Adjust design in inputs minimal

rejas:


eliasbraun (approved):
Looking Good!

packages/inputs-minimal/package.json
```diff
@@ -29,10 +29,7 @@
     "lint:typescript": "../../scripts/check-typescript"
   },
   "peerDependencies": {
-    "@babel/runtime": "^7.11.2",
```
peterhass, line 32: you can't just change peer dependency versions. every change in peer dep versions are breaking changes and therefore need a major version bumb (not worth in this case i guess)

https://wiki.nebenan.de/display/AWSM/1st+Party+Packages#id-1stPartyPackages-Peerdependencies

PR #185, opened 2023-01-06
[CORE-11519] Update ecosystembar to new design

rejas:


rejas:
Question: Should we update the components package version to 8.0.0 so that updates in BIZ and ORG dont fetch this design change automatically?

packages/components/src/ecosystem_bar/hooks.tsx
```diff
@@ -1,9 +1,9 @@
-import { useEffect, useState } from 'react';
+import { RefObject, useEffect, useState } from 'react';
 import eventproxy from 'nebenan-eventproxy';
 import useStableCallback from 'nebenan-react-hocs/lib/use_stable_callback';
 import { useEventListener } from 'nebenan-react-hocs/lib/use_event_listener';
 
-export const useScrolled = (navRef) => {
+export const useScrolled = (navRef: RefObject<HTMLElement>) => {
   const [scrolled, setScrolled] = useState(false);
```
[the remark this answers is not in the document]
rejas: Yes we can

packages/components/src/ecosystem_bar/index.tsx
```diff
@@ -1,23 +1,39 @@
...
+};
+
+export type EcosystemBarProps = {
+  items: EcosystemBarItem[];
+  onFirstSwipe: () => void;
+  onItemClick: (event: string) => void
+};
+
+export const EcosystemBar = ({ items, onFirstSwipe, onItemClick }: EcosystemBarProps) => {
   const navRef = useRef(null);
```
peterhass, line 24: type me

packages/components/src/ecosystem_bar/index.tsx
```diff
@@ -1,23 +1,39 @@
...
+    mobile: string,
+    others: string,
+  },
+};
+
+export type EcosystemBarProps = {
+  items: EcosystemBarItem[];
+  onFirstSwipe: () => void;
+  onItemClick: (event: string) => void
+};
```
eliasbraun, line 21: Just for my understanding: why do we export the EcosystemBarProps here?

packages/components/src/ecosystem_bar/index.tsx
```diff
@@ -1,23 +1,39 @@
...
+
+export const EcosystemBar = ({ items, onFirstSwipe, onItemClick }: EcosystemBarProps) => {
+  const navRef = useRef<HTMLUListElement>(null);
   const scrolled = useScrolled(navRef);
 
   useOnceSwipeTracking(navRef, onFirstSwipe);
 
-  const renderItem = (item) => (
+  const renderItem = (item:EcosystemBarItem) => (
```
eliasbraun, line 29: Maybe insert a space after item?

eliasbraun (approved):

packages/components/src/ecosystem_bar/index.tsx
rejas: typed

packages/components/src/ecosystem_bar/index.tsx
rejas: Good question, I was just doing it like the other components with the assumption so that it can be found when using the component in our apps. but then again I am new to this too :-) @peterhass  ?

packages/components/src/ecosystem_bar/index.tsx
rejas: will do. will also look if this https://typescript-eslint.io/rules/type-annotation-spacing/ is set or not...

packages/components/src/ecosystem_bar/index.tsx
eliasbraun: My assumption would be: I think there is no need in this case to export this props because they are bound to EcosystemBar and not really needed somewhere else. 
As I understood it: for more general types that might be needed somewhere else we can export those in a index.d.ts file on an upper root level [like here](https://github.com/good-hood-gmbh/nebenan-frontend/blob/fbe013e0d506f7a29ba0f9de965e3a1a52b3209f/client/types.tsx#L4) e.g.

rejas: True. But I will instead export the EcosystemBarItem type since this will be used outside in the nebenan-frontend https://github.com/good-hood-gmbh/nebenan-frontend/blob/e0c586260179963722a50ed5f7a6d89dbd2769cd/client/components/layout/ecosystem_nav_bar/utils.jsx#L4

packages/components/src/ecosystem_bar/index.tsx
```diff
@@ -1,23 +1,39 @@
...
+  link: string,
+  isActive: boolean,
+  event: string,
+  text: {
+    mobile: string,
+    others: string,
+  },
+};
+
+export type EcosystemBarProps = {
```
peterhass, line 17: I would rather start out with not exporting `EcosystembarProps` and `EcosystemBarItem`. We don't really know if we need them separately and if we start exporting them now, then we need a major release to remove them (breaking change).

packages/components/package.json
```diff
@@ -25,40 +26,27 @@
...
+    "@goodhood/icons": "^3.5.0",
+    "@popperjs/core": "^2.11.6",
+    "clsx": "^1.2.1",
+    "lodash": "^4.17.21",
     "escape-html": "^1.0.3",
-    "marked": "^4.0.0",
-    "nebenan-react-hocs": "^9.0.0"
+    "marked": "^4.2.5",
+    "nebenan-react-hocs": "^9.0.0",
+    "react-load-script": "^0.0.6"
```
peterhass, line 44: this needs to stay a peer dep (it has a global register of scripts). maybe you accidentally added it here instead of devDependencies?

packages/components/package.json
```diff
@@ -25,40 +26,27 @@
...
-    "prop-types": "^15.7.2",
-    "qs": "^6.9.4",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
-    "react-load-script": "^0.0.6",
     "react-redux": "^7.2.2",
-    "react-router": "^5.2.0",
-    "react-router-dom": "^5.2.0",
-    "redux": "^4.0.5",
-    "sass-functions": "^2.0.0",
```
peterhass, line 45: Are you really sure we can remove all those devDeps? I think most of them are deps of peerDependencies - `npm i` doesn't cry?

packages/components/package.json
```diff
@@ -14,6 +14,7 @@
   "version": "7.5.2",
```
peterhass, line 14: 7.6.0?

packages/components/index.d.ts
```diff
@@ -2,3 +2,30 @@ declare module '*.scss' {
...
+    callback: () => void
+  ): () => void;
+
+  export default eventproxy;
+}
+
+declare module 'nebenan-react-hocs/lib/use_stable_callback' {
+  function useStableCallback(
+    callback: () => void
+  ) : () => void;
```
peterhass, line 18: https://github.com/good-hood-gmbh/nebenan-frontend/blob/a76a106715a3ade3050cc25bfe93d1092d7ee93f/index.d.ts#L161

packages/components/index.d.ts
```diff
@@ -2,3 +2,30 @@ declare module '*.scss' {
...
+  export default useStableCallback;
+}
+
+declare module 'nebenan-react-hocs/lib/use_event_listener' {
+  import { RefObject } from 'react';
+
+  export function useEventListener<Type extends keyof HTMLElementEventMap>(
+    ref: RefObject<HTMLElement>,
+    type: Type,
+    handler: () => void,
```
peterhass, line 29: handlers do receive arguments - we might need to work on that definition here

event data should be available in `HTMLElementEventMap`. should work with some typescript magic

peterhass (changes requested):

packages/components/src/ecosystem_bar/index.tsx
rejas: will do

packages/components/package.json
rejas: Still stays as a peer dep, i just moved it into deps from devdeps since it is used in advertimesnt component. anyway, will undo the package updates to get this PR forward

packages/components/package.json
rejas: will do the package updates later when we get typescript export ready

packages/components/index.d.ts
rejas: In the linked code <T> is a generic type? Shouldnt it say somehow that callback a function?

peterhass: Ah, you're right! You could extend the generic `T` to something 'callable' (don't know how to do it myself) 

I think main point of the declaration is that whatever callback you pass in, type of returned callback is the same

PR #190, opened 2023-01-19
Update icons package to React18

rejas:
Had some breakfast cleanups for the icons lying around so why not push those and update the React version too?


rejas:
weird, now I get this when updating the lock file locally:

error An unexpected error occurred: "expected workspace package to exist for \"csso\"".

rejas:
Fixed error by downgrading to yarn 1.19.1 

Yarn 1 doesnt seem to get bugfixes anymore, maybe (sooner or later) time to move away from v1 and to v3 (or any other monorepo solution (lerna / nx / whatever))

peterhass:
Issue with yarn 2 is that their node_modules alternative is still not fully working. I tried it in the at layout toolchain PR and it still produces soo many issues you have to work around

packages/icons/CHANGELOG.md
```diff
@@ -25,3 +25,16 @@ sed -i '' 's#@goodhood/icons/lib/tiny/arrow_up#@goodhood/icons/lib/tiny/chevron_
 sed -i '' 's#@goodhood/icons/lib/tiny/arrow_left#@goodhood/icons/lib/tiny/chevron_left#g' **/*.jsx
 sed -i '' 's#@goodhood/icons/lib/tiny/arrow_right#@goodhood/icons/lib/tiny/chevron_right#g' **/*.jsx
 ```
+
+## v4.0.0 Changes
```
peterhass, line 29: I really like the trend of providing ready-to-use migration scripts.

Maybe make it work with `tsx` files as well?

packages/icons/package.json
```diff
@@ -25,25 +25,24 @@
...
+    "svgo": "^3.0.2",
+    "svgo-loader": "^3.0.3",
     "val-loader": "^2.1.2",
     "webpack": "^5.36.2",
     "webpack-cli": "^4.2.0",
     "webpack-dev-server": "^3.11.3"
   },
   "peerDependencies": {
-    "react": "^16.14.0 || ^17.0.1"
+    "react": "^16.14.0 || ^17.0.1 || ^18.2.0"
```
peterhass, line 46: maybe we just go for `>= 16.14.0` and hope for the best?

peterhass (approved):

rejas:
> Issue with yarn 2 is that their node_modules alternative is still not fully working. I tried it in the at layout toolchain PR and it still produces soo many issues you have to work around

Yarn 3 is already out, maybe that'll work?

packages/icons/package.json
rejas: Don't have enough expertise with the peerDependencies for this. What are your thoughts @eliasbraun @iamuchejude ?

peterhass: `>= 16.14.0` wouldn't break compatibility but it pretty much says "if you're lucky react 19 maybe works"

packages/icons/CHANGELOG.md
rejas: Done

packages/icons/package.json
rejas: Then lets try our luck :-)

PR #191, opened 2023-01-31
Cleanup dependencies

rejas:
to get rid of that lodash monolith import in core

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "7.6.1",
+  "version": "7.6.2-beta.0",
   "module": "lib/index.esm.js",
   "main": "lib/index.js",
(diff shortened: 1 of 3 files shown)


packages/components/package.json
```diff
@@ -26,34 +26,21 @@
...
     "nebenan-keymanager": "^1.0.1",
-    "nebenan-ui-kit": "^5.1.0",
-    "node-sass-glob-importer": "^5.3.2",
-    "prop-types": "^15.7.2",
-    "qs": "^6.9.4",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
     "react-load-script": "^0.0.6",
     "react-redux": "^7.2.2",
-    "react-router": "^5.2.0",
```
peterhass, line 43: are you sure we don't need those? i think we're using react-router in preview

rejas: those devDependencies are all listed in the root package-json

peterhass (approved):

PR #195, opened 2023-02-07
TypeScript support for @goodhood/chargebee 

iamuchejude:


packages/chargebee/src/checkout/index.tsx
```diff
@@ -22,22 +30,13 @@ const Checkout = ({
       success: () => { isSuccess = true; },
       error: onError,
       close: () => {
-        invoke(onClose);
```
peterhass, line 25: `invoke(onClose)` is not functionally equal to `onClose?.()`

Wouldn't do a breaking release because of it but just watch out when testing in apps

iamuchejude: Good point. I would just a `.call` instead of just invoking. That should replace the `is function` guard clause.

packages/chargebee/index.d.ts
```diff
@@ -2,3 +2,26 @@ declare module '*.scss' {
...
+    onError?: () => void;
+    onCreate?: () => void;
+    attributes?: Omit<
+    React.HTMLAttributes<HTMLScriptElement>,
+    'onload' | 'onerror' | 'src'
+    >;
+  };
+
+  const Script: React.ComponentType<ScriptProps>;
+  export default DefaultIcon;
```
peterhass, line 20: this doesn't sound right.

packages/chargebee/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+  enableRedirectMode?: boolean;
+  enableFriendbuyTracking?: boolean;
+  enableRefersionTracking?: boolean;
+};
+
+type Chargebee = {
+  init: (options: ChargebeeInitOptions) => ChargebeeInstance;
+  getInstance: () => ChargebeeInstance;
+
+  getPortalSections: () => Record<string, string>;
```
peterhass, line 137: https://www.chargebee.com/checkout-portal-docs/api.html#getportalsections

packages/chargebee/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+  token: PortalSessionToken;
+  status: PortalSessionStatus;
+  object: 'portal_session';
+  customer_id: CustomerId;
+  access_url: string,
+  created_at: number,
+  expires_at: number,
+};
+
+export type PortalSessionSetter = () => Promise<PortalSession>;
```
peterhass, line 101: isn't that actually some kind of getter?

peterhass: getSession?

packages/chargebee/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+};
+
+type ChargebeePortal = {
+  open: (
+    options: ChargebeePortalOptions,
+    forwardOptions?: ChargebeeForwardOptions
+  ) => void;
+};
+
+type CustomerId = Nominal<string, 'CustomerId'>;
```
peterhass, line 79: I think we need to export those nominal types, otherwise you can't set those props outside of this package

packages/chargebee/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+  step?: (currentStep: string) => void;
+};
+
+type ChargebeePortalOptions = {
+  close?: VoidFunction;
+  paymentSourceAdd?: VoidFunction;
+  paymentSourceUpdate?: VoidFunction;
+  paymentSourceRemove?: VoidFunction;
+  loaded?: VoidFunction;
+  visit?: (sectionType: string) => void;
```
peterhass, line 58: pretty sure that this section type matches the ones listed here

https://www.chargebee.com/checkout-portal-docs/api.html#getportalsections

packages/chargebee/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+  | 'full_page';
+
+export type OpenCheckoutOptions = {
+  hostedPage: () => HostedPage;
+  close?: VoidFunction;
+  success: (hostedPageId: HostedPageId) => void;
+  error?: VoidFunction;
+  loaded?: VoidFunction;
+  layout?: OpenCheckoutLayout;
+  step?: (currentStep: string) => void;
```
peterhass, line 49: any way to type this string? doesn't sound like it's dynamic in any way

packages/chargebee/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+  | 'activated';
+
+type PortalSession = {
+  id: PortalSessionId;
+  token: PortalSessionToken;
+  status: PortalSessionStatus;
+  object: 'portal_session';
+  customer_id: CustomerId;
+  access_url: string,
+  created_at: number,
```
peterhass, line 97: are you sure all those dates are unix timestamps? really wonder why they don't send ISO strings

packages/chargebee/src/types.ts
iamuchejude: Missed that. Coming soon.

packages/chargebee/src/types.ts
iamuchejude: Of course, we need to export then.

packages/chargebee/index.d.ts
iamuchejude: At all.

packages/chargebee/src/types.ts
iamuchejude: I also think it's more of a getter even though their documentation calls it a `setterFunction`.

packages/chargebee/src/types.ts
iamuchejude: 🤔  So far, they only send unix timestamps.

packages/chargebee/src/types.ts
iamuchejude: Are you suggesting narrowing the type?

rejas (approved):

packages/chargebee/src/portal_link/index.tsx
```diff
@@ -1,19 +1,34 @@
...
-}) => {
-  const handleOpenLink = (instance, Chargebee) => {
+}: PortalLinkProps) => {
+  const handleOpenLink: OnCallHandler = (instance, Chargebee) => {
     let forwardOptions;
 
     if (section) {
-      forwardOptions = { sectionType: Chargebee.getPortalSections()[section] };
+      forwardOptions = {
+        sectionType: Chargebee?.getPortalSections()[section],
```
peterhass, line 30: any idea in which case `Chargebee` can be undefined here?

packages/chargebee/src/portal_link/index.tsx
```diff
@@ -1,19 +1,34 @@
...
+  OnCallHandler,
+} from '../types';
 import Action from '../action';
 
+type PortalLinkProps = {
+  site: string;
+  onSessionGet: GetPortalSessionFunction;
+  onClose?: () => void;
+  section?: PortalSectionsKeys;
+} & Record<string, unknown>;
```
peterhass, line 17: Use `Action` props here? You're passing them through

packages/chargebee/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "1.5.1",
+  "version": "1.6.0",
```
rejas, line 14: you can get rid of the proptypes devDependency since it is already imported via the root package.json

packages/chargebee/src/portal_link/index.tsx
```diff
@@ -1,19 +1,34 @@
...
+import {
+  ChargebeeForwardOptions,
+  PortalSectionsKeys,
+  GetPortalSessionFunction,
+  OnCallHandler,
+  VoidFunction,
+} from '../types';
+import Action, { ActionProps } from '../action';
+
+type PortalLinkProps = ActionProps & {
```
peterhass, line 13:
```suggestion
type PortalLinkProps = Omit<ActionProps, 'section' | 'onSessionGet' | 'onClose'> & {
```

peterhass (approved):

packages/chargebee/src/action/index.tsx
```diff
@@ -32,12 +37,18 @@ const Action = ({
...
+        onCall?.call?.(
+          undefined,
+          instanceRef.current as ChargebeeInstance,
+          global.Chargebee,
+        );
+      }
     };
 
-    node = <span {...rest} onClick={handleClick} />;
+    node = <span {...rest} role="button" tabIndex={0} onClick={handleClick} />;
```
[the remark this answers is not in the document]
iamuchejude: I don't think it's a problem with this PR. `span` is not a `button` but we need a button there and it should be focusable. I was gonna switch the tag to button but that comes with some extra work.

It's really just a small improvement

packages/chargebee/src/portal_link/index.tsx
iamuchejude: Why omit when `ActionProps` doesn't include any of those values?

packages/chargebee/src/action/index.tsx
```diff
@@ -32,12 +37,18 @@ const Action = ({
...
   if (isReady) {
-    const handleClick = (event) => {
-      invoke(onClick, event);
-      if (!disabled) invoke(onCall, instanceRef.current, global.Chargebee);
+    const handleClick: React.MouseEventHandler<HTMLSpanElement> = (event) => {
+      onClick?.call?.(undefined, event);
+      if (!disabled) {
+        onCall?.call?.(
+          undefined,
+          instanceRef.current as ChargebeeInstance,
```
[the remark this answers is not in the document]
iamuchejude: The pattern doesn't sit right with me too but I think type assertion here is inevitable especially because the value here (`ref`) can be more than 1 type. We are already accounting for the null/undefined value, so it's safe to assume that the value is not null/undefined here

packages/chargebee/src/checkout/index.tsx
```diff
@@ -0,0 +1,41 @@
+import Action, { ActionProps } from '../action';
+import { OpenCheckoutOptions, OnCallHandler, VoidFunction, HostedPage } from '../types';
+
+type CheckoutProps = ActionProps & {
```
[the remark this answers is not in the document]
peterhass: Because we set it ourself inside of the component so it should be excluded from the type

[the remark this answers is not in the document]
peterhass: Down here you're passing `onCall` https://github.com/goodhood-eu/goodhood/pull/195/files/2374c4a247913625a3dfc50b185ba3f664ceafe7#diff-3a88d6e20c4a6ee012aa90a65c094b89eea3403f57c373f596a0dca48a776c8dR38

Users of `Checkout` shouldn't be able to set `onCall`, therefore it shouldn't be in the type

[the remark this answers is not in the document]
iamuchejude: Oh, now I see it. Took me forever though 🤔

packages/chargebee/src/portal_link/index.tsx
peterhass: ok seems like now you only need to omit `onCall`

PR #197, opened 2023-02-21
[CORE-11326] @goodhood/tracking: React 18 support

eliasbraun:
Nothing really to add here. Is it a non breaking change then?

### DIFF
--- a/packages/tracking/package.json
+++ b/packages/tracking/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "2.0.0",
+  "version": "1.1.0",
   "module": "lib/index.esm.js",
   "main": "lib/index.js",
   "sideEffects": false,
--- a/packages/tracking/package.json
+++ b/packages/tracking/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "1.0.2",
+  "version": "2.0.0",
   "module": "lib/index.esm.js",
   "main": "lib/index.js",
   "sideEffects": false,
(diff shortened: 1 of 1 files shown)


packages/tracking/package.json
```diff
@@ -28,10 +28,11 @@
     "lint:typescript": "../../scripts/check-typescript"
   },
   "devDependencies": {
-    "prop-types": "^15.8.1"
+    "prop-types": "^15.8.1",
+    "nebenan-helpers": "^6.0.0"
```
rejas, line 32: why do we need this in devDependencies?

eliasbraun: Added it here, because VSCode was shouting at me that this is missing in deps :)

rejas: But its in the root package.json.... Webstorm doesnt complain :-(

eliasbraun: Hmm true! Will get rid of it again!

packages/tracking/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "1.0.2",
+  "version": "2.0.0",
```
eliasbraun, line 14: What do you think?

rejas: Probably indeed a v2 since I remember peters mantra: Updated peerDependencies === Updated Major Version :-)

peterhass: This PR is **adding** support for react 18, so it's not a breaking change -> therefore it's 1.1.0

peterhass: Wrote a small section on the wiki about it 

https://wiki.nebenan.de/display/AWSM/1st+Party+Packages#id-1stPartyPackages-Situation2.1:Addingsupportformoreversions

peterhass (changes requested):
-> minor version bump

rejas (approved):

peterhass (approved):
Looks good but how did you test it if our preview is still running react 16?

https://github.com/goodhood-eu/goodhood/blob/master/package.json#L70

eliasbraun:
Already solved in
https://github.com/goodhood-eu/goodhood/pull/198

PR #198, opened 2023-02-28
Preview: Update to react 18

peterhass:


eliasbraun (approved):
tested it, looks good and works on my machine!

PR #199, opened 2023-03-13
Components: Typescriptify Markdown

peterhass:


packages/components/src/index.tsx
```diff
@@ -0,0 +1,43 @@
+export { default as Advertisement } from './advertisement';
```
peterhass, line 1: needed to write all those exports manually because the glob-expand rollup plugin we used there doesn't work with typescript

packages/components/src/markdown/index.tsx
```diff
@@ -1,22 +1,34 @@
...
 import { sanitizeText } from './utils';
-import { useClickHandler } from './hooks';
+import { ClickHandler, useClickHandler } from './hooks';
+import { ComponentPropsWithoutRef, PropsWithChildren } from 'react';
 
-const options = { gfm: false };
+const options = { gfm: false } as const;
+
+type RemainingHTMLAttributes = Omit<ComponentPropsWithoutRef<'span'>,
+'className' | 'onClick'
```
rejas, line 12: are we omiting them here to be more explicit that className is allowed as a Prop or why is it?

packages/components/src/index.tsx
rejas: is there an issue at the rollup plugin maybe that we can track?

peterhass: Didn't find one. We would need to do a PR on their side

rejas: ah, i see: https://github.com/gjbkz/rollup-plugin-glob-import/issues/364

packages/components/src/markdown/index.tsx
peterhass: type includes all props a span would have **but className and onClick**

rejas: I see, but why?

peterhass: because we're setting them in line 38

peterhass: https://github.com/goodhood-eu/goodhood/blob/32edd4448837dec08ae252bf2922f9175c23d137/packages/components/src/markdown/index.tsx#L38

PR #201, opened 2023-04-19
[CORE-11894] Add support of data-testId in ecosystem bar items

spossner:
[JIRA Link](https://jira.nebenan.de/browse/CORE-11894)

In order to make the items in the ecosystem bar accessible via data-testid selectors, we need to support such data attributes in the specification and rendering of the ecosystem navbar items.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "7.7.0",
+  "version": "7.7.1",
   "module": "lib/index.esm.js",
   "main": "lib/index.js",
   "types": "lib/types/packages/components/src",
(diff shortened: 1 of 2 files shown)


eliasbraun (approved):

packages/components/package.json
```diff
@@ -11,7 +11,7 @@
   "bugs": {
     "url": "https://github.com/goodhood-eu/goodhood/issues"
   },
-  "version": "7.7.0",
+  "version": "7.7.1",
```
eliasbraun, line 14: I think it's technically not a bugfix, so one could argue that this could be 7.8.0 Version bump? No strong opinion from my side tho

peterhass, line 14: whatever Elias says!

peterhass (changes requested):

packages/components/package.json
spossner: will change to 7.8.0 -> good point 🙏

spossner: dito! Thanks!

PR #204, opened 2023-04-20
Update @goodhood/components to react 18

spossner:
[JIRA Link](https://jira.nebenan.de/browse/CORE-11326)

Update packages/components to support React 18

### DIFF
--- a/packages/components/src/ecosystem_bar/index.stories.jsx
+++ b/packages/components/src/ecosystem_bar/index.stories.jsx
@@ -1,4 +1,4 @@
-import EcosystemBar from '@/src/ecosystem_bar/index';
+import EcosystemBar from '.';
 
 export default { title: 'EcosystemBar', component: EcosystemBar };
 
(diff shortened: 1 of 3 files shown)


spossner:
HELP
I can not yarn start on my machine with yarn 1.22.19, node 16.19.1
Hickup and conflicts with icons
Compiles and preview starts on Elias' machine.
🤷‍♂️

spossner:
succeeded build and preview on linux system at home... -> fine for review + merge + publish from my side

packages/components/src/ecosystem_bar/index.stories.jsx
```diff
@@ -1,4 +1,4 @@
-import EcosystemBar from '@/src/ecosystem_bar/index';
+import EcosystemBar from '.';
```
spossner, line 1: this seems to not work in the monorepo type of setup..

eliasbraun: Interesting... Does 
`import EcosystemBar from './index'; `
work for you? That would be the usual way of importing, maybe we could add it here to keep it consistent?

eliasbraun (approved):

PR #207, opened 2023-06-12
Fix build-preview task

rejas:
was broken due to a ts helper function (__rest) being imported twice when used in another monorepo package The error happenend in modals package which was the only package using @goodhood/components.

This PR changes the ts compile target to esnext (therefore removing the need for the __rest helper function) and lets babel transpile to whatever we need according to our browserslist.

But this might only cure the symptoms, not the root cause of the error. So maybe in the future something else like this can happen (due to a helper function being needed and being imported twice). So you can blame me future-veeck :-)


packages/components/index.d.ts
```diff
@@ -33,7 +33,7 @@ declare module 'nebenan-react-hocs/lib/use_event_listener' {
 declare module 'nebenan-helpers/lib/utils' {
   export function invoke<TFunc>(
     fn: TFunc | undefined | null,
-    ...args: Paramters<TFunc>
+    ...args: Parameters<TFunc>
```
iamuchejude, line 36: Interesting how our linter and compiler didn't catch this 🤔

iamuchejude (approved):
🦸🏿

packages/components/index.d.ts
rejas: looks like our current linting via "tsc --noEmit" doesn't catch errors in the index definition files at all. Try out by adding something stupid there in core -> nothing....

Maybe we have to switch to a "proper" linter sometime in the future?

iamuchejude: `tsc --noEmit` is mostly used for type-checking without outputting transpiled code 🤔 

There's a card on Favro about linters, so, we can discuss it then.

PR #209, opened 2023-07-04
[CORE-13066] implement GA4 in tracking lib

fweinrich:


packages/tracking/package.json
```diff
@@ -30,5 +30,8 @@
   "peerDependencies": {
     "@babel/runtime": "^7.12.5",
     "react": ">= 16.14.0"
+  },
+  "dependencies": {
+    "react-load-script": "^0.0.6"
```
rejas, line 35: one of those old libraries, that are not maintained anymore: https://www.npmjs.com/package/react-load-script

maybe time to either look for a replacement or write a little helper ourselves?

rejas (approved):

packages/tracking/src/provider.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+    selector: /\/feed\/marketplace$/,
+    section: 'marketplace',
+    track: {
+      page_name: 'marketplace_feed',
+    },
+  },
+];
+export const Default = () => (
+  <Provider
+    enableAnalytics gtmId="GTM-56G85MT" baseEvent={{
```
spossner, line 38: credentials into @root/config

packages/tracking/src/provider.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <AnalyticsProvider value={context}>
+      <PageView
+        enabled={enableAnalytics && enablePageTracking}
+        pageMapping={pageMapping}
+      >
+        {enableAnalytics && (
+          <Script
+            url={getScriptSource(gtmId)} onCreate={setup}
```
spossner, line 37: can be done async with
`attributes={{ async: true }}`

packages/consent/src/index.ts
```diff
@@ -0,0 +1,11 @@
+import { loadPrivacyManagerModal } from './sourcepoint';
```
iamuchejude, line 1: We'd need the consent library to be app and vendor-list agnostic. And that makes type-safety of the APIs more complex to achieve especially with `useGrant` hooks. Also, we have those setups/configurations that need to happen for consent to work

All of these are mostly speculation but, I think we should take more time for the consent module, maybe have a separate ticket and do more design discussion before we move to it. I think we can keep the in-place consent module pending when this is ready,

packages/tracking/src/types.ts
```diff
@@ -0,0 +1,167 @@
...
+  current_plan: string;
+  currency: string;
+  value: string;
+  item_name: string;
+  transaction_id: string;
+  coupon: string;
+  payment_type: string;
+};
+
+export type PageSection =
```
iamuchejude, line 27: Are these the same for all our apps?

iamuchejude (commented):
👓

packages/tracking/src/types.ts
fweinrich: yes! since they are used in the GA reports gobally i wanted to type them

packages/tracking/src/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+  } else {
+    console.log(item);
+  }
+};
+
+type LoggerOptions = {
+  collapsed?: boolean;
+};
+
+const getLogger = (id: string, options: LoggerOptions = {}) => (...args: unknown[]) => {
```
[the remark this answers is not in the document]
fweinrich: I moved it to @goodhood/helpers (see https://github.com/goodhood-eu/goodhood/pull/210 ) since it feels bad to remove all the typings. Let's create a new ticket for migrating nebenan-helpers into @goodhood/helpers. It's the first step for unifying all of our packages!

packages/consent/src/index.ts
fweinrich: Let's create tickets for fully migrating consent to @goodhood/consent!

packages/tracking/src/hooks/use_track.ts
```diff
@@ -0,0 +1,66 @@
...
+      log('tracking without dataLayer', event, payload);
+    } else {
+      log('tracking to dataLayer', event, payload);
+      window.dataLayer.push({
+        event,
+        section: match?.section,
+        ...baseEvent,
+        ...payload,
+        analytics_storage: hasAnalyticsStorageConsent,
+        ad_storage: hasAnalyticsStorageConsent,
```
fweinrich, line 62: should be removed since we are only doing analytics here

packages/tracking/src/context.ts
```diff
@@ -0,0 +1,17 @@
...
+import { BaseEvent, PageMapping } from './types';
+
+export type TAnalyticsContext = {
+  baseEvent?: BaseEvent,
+  pageMapping: PageMapping[],
+  hasAnalyticsStorageConsent: boolean,
+  hasGoogleTagManagerConsent: boolean
+};
+
+export const AnalyticsContext = createContext<TAnalyticsContext>({
```
fweinrich, line 11: rename to trackingContext

packages/tracking/src/hooks/use_track.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export type TrackFunction =
+    <K extends keyof TrackEventMap>(event: K, payload: TrackEventMap[K]) => void;
+
+
+export const useTrack = (): TrackFunction => {
+  const { baseEvent, pageMapping, hasAnalyticsStorageConsent } = useAnalytics();
+  const location = useLocation();
+  const match = useMemo(
```
fweinrich, line 47: rename match to PageMapping

packages/tracking/src/hooks/use_track_page_view.ts
```diff
@@ -0,0 +1,19 @@
...
+import { useAnalytics, useTrack } from './use_track';
+
+
+export const useTrackPageView = () => {
+  const location = useLocation();
+  const track = useTrack();
+  const { pageMapping } = useAnalytics();
+
+  useEffect(() => {
+    const match = pageMapping.find((m) => m.selector.test(location.pathname));
```
fweinrich, line 12: rename match to PageMapping

packages/tracking/src/types.ts
```diff
@@ -0,0 +1,114 @@
...
+export type SwipeEvent = InteractionEvent & ContentEvent & {
+  swipe_direction: string;
+};
+
+export type ViewEvent = InteractionEvent & ContentEvent & {
+  search_term: string;
+  content_position: string;
+};
+
+export type PageMapping = {
```
fweinrich, line 88: rename to PageMap

PR #220, opened 2023-09-18
Fix type definition in modal package

rejas:
locked and inverted can be optional since they have default values


iamuchejude (approved):
LGTM

PR #222, opened 2023-09-27
Make filter_category type optional

rejas:
fixes typescript warnings in our core ts checks


fweinrich (approved):
lgtm, will release now

